PR #3, opened 2015-11-18
Incomplete bookshelf sample.

SurferJeffAtGoogle:
Uses CloudStorage and Cloudsql.  All tests pass and code coverage is
100% in src/.


SurferJeffAtGoogle:
Brent, I accommodated all your comments from pull #1.

.gitignore
```diff
@@ -1,2 +1,6 @@
 **/vendor/
 **/coverage/
+.idea
+*~
+*.iml
+composer.phar
```
JustinBeckwith: do we need to have license headers on config files?

bookshelf/src/CloudsqlModel.php
```diff
@@ -0,0 +1,163 @@
...
+class CloudsqlModel implements DataModel
+{
+    /**
+     * Creates a new PDO instance and sets error mode to exception.
+     *
+     * @return PDO
+     */
+    private function newConnection()
+    {
+        $pdo = new PDO(getenv('MYSQL_DSN'), getenv('MYSQL_USER'),
```
JustinBeckwith: Should we use something like .env for these?

JustinBeckwith:
LGTM

bshaffer:
all twig HTML files should have the extension `.html.twig`

bookshelf/templates/view.html
```diff
@@ -0,0 +1,55 @@
...
+  </div>
+</form>
+
+<div class="media">
+  {# [START book_image] #}
+  <div class="media-left">
+    {% if book.imageUrl %}
+      <img class="book-image" src="{{book.imageUrl}}">
+    {% else %}
+      <img class="book-image" src="http://placekitten.com/g/128/192">
```
bshaffer: I would rather see this:

```html
<img class="book-image" src="{{ book.imageUrl ?: 'http://placekitten.com/g/128/192' }}">
```

Or even better, set the default to a var, so it's `book.imageUrl ?: defaultImage`

bshaffer:
we should add a `phpunit.xml` which points to a `bootstrap.php` file, so we don't have to include the composer autoloader in every test. Then the tests can truly be stand alone, and also we can load other test classes/etc in our bootstrap file if we need to.

Also, all trait names should end with `Trait`, and all interfaces should end in `Interface`, so we need to rename `DataModel` to `DataModelInterface`, and `DataModelTest` to `DataModelTestTrait`

In the same vein, we should move `DataModelTestTrait` into a `test/lib` directory, and use `require_once` in `bootstrap.php`

tmatsuo:
I don't see `app.yaml` in the changes. Did you successfully deploy it?

SurferJeffAtGoogle:
No app.yaml yet.  I'll do that in a later PR.

.gitignore
SurferJeffAtGoogle: I don't think so.  I haven't seen that pattern in our other repositories.

bookshelf/src/CloudsqlModel.php
SurferJeffAtGoogle: When running tests on Travis, Jenkins, etc., it's very useful to have these in environment variables, so that there are no conflicts when merging between a base and a fork.

SurferJeffAtGoogle:
Ready for another look.

bookshelf/src/CloudsqlModel.php
```diff
@@ -0,0 +1,163 @@
...
+
+        return array(
+            'books' => $rows,
+            'cursor' => $new_cursor,
+        );
+    }
+
+    public function create($book, $id = null)
+    {
+        $names = '';
```
bshaffer: If you want to check for invalid properties, one thing you could do is:

```php
```

bookshelf/src/CloudsqlModel.php
```diff
@@ -0,0 +1,163 @@
...
+        $values = [];
+        $question_marks = '';
+        $separator = '';
+        $pdo = $this->newConnection();
+        foreach ($this->columnNames as $column) {
+            if (array_key_exists($column, $book)) {
+                $question_marks = "$question_marks$separator?";
+                $names = "$names$separator$column";
+                $separator = ', ';
+                array_push($values, $book[$column]);
```
bshaffer: For these, I would prefer to see

```php
foreach(...) {
    if (...) {
        $question_marks[] = '?';
        $names[] = $column
        $values[] = $book[$column];
    }
}
$sql = sprintf('INSERT INTO books (%s) VALUES(%s)',
    implode(', ', $names),
    implode(', ', $query_marks)
);
//..
```

Or even better, avoid the `foreach` all together and do something like this:

```php
$names = array_keys($books);
$question_marks = array_fill(0, count($books), '?');
$sql = sprintf('INSERT INTO books (%s) VALUES(%s)',
    implode(', ', $names),
    implode(', ', $query_marks)
);
$statement = $pdo->prepare($sql);
$statement->execute($book);
```

bookshelf/src/CloudsqlModel.php
```diff
@@ -0,0 +1,163 @@
...
+        $statement->execute($values);
+
+        return $statement->rowCount();
+    }
+
+    public function delete($id)
+    {
+        $pdo = $this->newConnection();
+        $statement = $pdo->prepare('DELETE FROM books WHERE id = ?');
+        $statement->bindValue(1, $id, PDO::PARAM_INT);
```
bshaffer: You can also use strings to associate values, which is a little more readable:

```php
$statement = $pdo->prepare('DELETE FROM books WHERE id = :id');
$statement->bindValue('id', $id, PDO::PARAM_INT);
```

bshaffer: Ahh, I just noticed you're doing this in the above `read` function! We should be consistent here.

bookshelf/src/CloudsqlModel.php
```diff
@@ -0,0 +1,163 @@
...
+        foreach ($this->columnNames as $column) {
+            if (array_key_exists($column, $book)) {
+                $assignments = "$assignments$separator$column=?";
+                $separator = ', ';
+                array_push($values, $book[$column]);
+            }
+        }
+        $statement = $pdo->prepare(
+            "UPDATE books SET $assignments WHERE id = ?"
+        );
```
bshaffer: Same thing here as in `create`, I would like to see `$values[] = ...` instead of `array_push`, and `implode` or `sprintf` instead of embedding variables inside double quotes. Even better yet would be to avoid the `foreach`.

bookshelf/src/controllers.php
```diff
@@ -0,0 +1,153 @@
...
+});
+
+$app->get('/books/{id}/edit', function ($id) use ($app) {
+    /** @var DataModelInterface $model */
+    $model = $app['bookshelf.model'];
+    $book = $model->read($id);
+    if (!$book) {
+        return new Response('', 404);
+    }
+    $book = $model->read($id);
```
bshaffer: duplicate line here

bookshelf/src/controllers.php
```diff
@@ -0,0 +1,153 @@
...
+        );
+    }
+    if ($model->update($book)) {
+        if ($oldImageUrl) {
+            $storage->deleteFile($oldImageUrl);
+        }
+
+        return $app->redirect("/books/$id");
+    } else {
+        return new Response('', 404);
```
bshaffer: We can leave the `else` off here and just return, since the other side of the conditional will have returned by now

```php
if (...) {
    return $app->redirect(...);
}

return new Response('', 404);
```

bookshelf/src/controllers.php
```diff
@@ -0,0 +1,153 @@
...
+            /** @var ImageStorage $storage */
+            $storage = $app['bookshelf.storage'];
+            $imageUrl = isset($book['imageUrl']) ? $book['imageUrl'] : null;
+            if ($imageUrl) {
+                $storage->deleteFile($imageUrl);
+            }
+
+            return $app->redirect('/books/', 303);
+        } else {
+            return new Response('', 404);
```
bshaffer: same comment on the return

bookshelf/templates/base.html.twig
```diff
@@ -0,0 +1,40 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#}
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <title>Bookshelf - Python on Google Cloud Platform</title>
```
bshaffer: change Python to PHP

bookshelf/test/functional/controllersTest.php
```diff
@@ -0,0 +1,205 @@
...
+                $more = $crawler->filter('a:contains("More")');
+                if (count($more)) {
+                    $crawler = $client->click($more->link());
+                } else {
+                    break;
+                }
+
+            }
+            $this->assertTrue($foundTreasureIsland);
+            $this->assertTrue($foundCatHat);
```
bshaffer: It seems like we should get rid of the `try` here, but if it's necessary, we should add a `$this->fail(..)` in the `catch`, otherwise the test will succeed even if an exception is thrown and the images are not found.

bookshelf/test/lib/DataModelTestTrait.php
```diff
@@ -0,0 +1,86 @@
...
+
+        // Try updating a book.
+        $breakfastBook['description'] = 'A really fun read.';
+        $model->update($breakfastBook);
+        $breakfastBookCopy = $model->read($breakfastId);
+        // And confirm it was correctly updated.
+        $this->assertEquals(
+            'A really fun read.',
+            $breakfastBookCopy['description']
+        );
```
bshaffer: Very nice and extensive test suites all around! Some suggestions for this test:

 - test adding a property that doesn't exist
 - test updating a property that doesn't exist
 - test updating a single property and leaving the rest out, verify the other properties are not *empty* after saving
 - test updating a property and setting it empty, and make sure it *is* empty after you save

This would require splitting this into multiple functions

bookshelf/src/CloudsqlModel.php
SurferJeffAtGoogle: I wanted to be more permissive.  Strictly checking for invalid properties means that things can break across schema changes.   But that's probably being too paranoid for this sample.  So, I'll update the code.

bookshelf/src/CloudsqlModel.php
SurferJeffAtGoogle: Done.

bookshelf/src/CloudsqlModel.php
SurferJeffAtGoogle: Done.

bookshelf/src/CloudsqlModel.php
SurferJeffAtGoogle: Done.

bookshelf/src/controllers.php
SurferJeffAtGoogle: Done.

bookshelf/src/controllers.php
SurferJeffAtGoogle: Done.

bookshelf/src/controllers.php
SurferJeffAtGoogle: Done.

bookshelf/test/functional/controllersTest.php
SurferJeffAtGoogle: There is no catch.  There is only a finally.  So tests should correctly fail in an exception is thrown.

bookshelf/templates/base.html.twig
SurferJeffAtGoogle: Done.

bookshelf/test/lib/DataModelTestTrait.php
SurferJeffAtGoogle: > test adding a property that doesn't exist
Done.

> test updating a property that doesn't exist
Done. 

> test updating a single property and leaving the rest out, verify the other properties are not empty after saving
These semantics will be a hassle to implement on nosql databases, and are never actually needed in the sample, so I changed the semantics so that missing properties on the book means that those properties get removed from the model.  That will be much simpler to implement for datastore and the like.

> test updating a property and setting it empty, and make sure it is empty after you save
Done.

SurferJeffAtGoogle:
Ready for another look.

bookshelf/phpunit.xml
```diff
@@ -0,0 +1,10 @@
+<phpunit
+        bootstrap="test/bootstrap.php">
+    <testsuites>
+        <testsuite name="AllTests">
+            <directory>test/functional</directory>
+            <directory>test/unit</directory>
+        </testsuite>
+    </testsuites>
+
+</phpunit>
```
tmatsuo: Tiny tiny nit: add newline at the end?

bookshelf/src/CloudsqlModel.php
```diff
@@ -0,0 +1,173 @@
...
+{
+    /**
+     * Creates a new PDO instance and sets error mode to exception.
+     *
+     * @return PDO
+     */
+    private function newConnection()
+    {
+        $pdo = new PDO(getenv('MYSQL_DSN'), getenv('MYSQL_USER'),
+            getenv('MYSQL_PASSWORD'));
```
tmatsuo: What error message you will see when those env vars not available?

bookshelf/src/CloudsqlModel.php
```diff
@@ -0,0 +1,173 @@
...
+    public function create($book, $id = null)
+    {
+        $this->verifyBook($book);
+        if ($id) {
+            $book['id'] = $id;
+        }
+        $pdo = $this->newConnection();
+        $names = array_keys($book);
+        $placeHolders = array_map(function ($key) { return ":$key";}, $names);
+        $sql = sprintf('INSERT INTO books (%s) VALUES(%s)',
```
tmatsuo: Put a space after VALUES

bookshelf/src/DataModelInterface.php
```diff
@@ -0,0 +1,75 @@
...
+     *               'cursor' => pass to next call to listBooks() to fetch
+     *               more books]
+     */
+    public function listBooks($limit = 10, $cursor = null);
+
+    /**
+     * Creates a new book in the data model.
+     *
+     * @param $book array  An associative array.
+     * @param null $id integer  The id, if known.
```
tmatsuo: Add @return ?

bookshelf/src/DatastoreModel.php
```diff
@@ -0,0 +1,84 @@
...
+class DatastoreModel implements DataModelInterface
+{
+    public function __construct($datastoreDatasetId = null)
+    {
+        if (!$datastoreDatasetId) {
+            $datastoreDatasetId = getenv('GOOGLE_PROJECT_ID');
+        }
+        $this->datasetId = $datastoreDatasetId;
+        $client = new \Google_Client();
+        $client->setScopes([
```
tmatsuo: I don't think we still need these scopes if it's running on Managed VMs. Do we still need them for other cases?

bookshelf/src/app.php
```diff
@@ -0,0 +1,36 @@
...
+ */
+use Silex\Application;
+use Silex\Provider\TwigServiceProvider;
+
+$app = new Application();
+
+$app->register(new TwigServiceProvider(), array(
+    'twig.path' => __DIR__.'/../templates',
+    'twig.options' => array(
+        'strict_variables' => false,
```
tmatsuo: Can you set strict_variables to true? I feel that using strict_variables = true is preferred development practice, but if the application depends on it, that's fine.

bookshelf/src/app.php
```diff
@@ -0,0 +1,36 @@
...
+    ),
+));
+$app['debug'] = true;
```
tmatsuo: Is it worthwhile to introduce a new envvar for 'debug' value?

bookshelf/src/controllers.php
```diff
@@ -0,0 +1,150 @@
...
+            $image->getMimeType()
+        );
+    }
+    if ($model->update($book)) {
+        if ($oldImageUrl) {
+            $storage->deleteFile($oldImageUrl);
+        }
+
+        return $app->redirect("/books/$id");
+    }
```
tmatsuo: When $model->update($book) fails, the new image will be orphaned. Can you add some cleanup code if it's easy, or add some comments explaining the edge case?

bookshelf/src/controllers.php
```diff
@@ -0,0 +1,150 @@
...
+        if ($book) {
+            $model->delete($id);
+            /** @var ImageStorage $storage */
+            $storage = $app['bookshelf.storage'];
+            $imageUrl = isset($book['imageUrl']) ? $book['imageUrl'] : null;
+            if ($imageUrl) {
+                $storage->deleteFile($imageUrl);
+            }
+
+            return $app->redirect('/books/', 303);
```
tmatsuo: Is it possible to use the constants in Symfony's Response class?

https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpFoundation/Response.php#L23

bookshelf/test/lib/DataModelTestTrait.php
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        // Clean up.
+        $model->delete($breakfastId);
+        $this->assertFalse($model->read($breakfastId));
+        $this->assertTrue((bool)$model->read($bellId));
+        $model->delete($bellId);
+        $this->assertFalse($model->read($bellId));
+    }
+}
```
tmatsuo: Nit: Add newline at the end

tmatsuo:
@SurferJeffAtGoogle I took a look and added some comments.

bookshelf/src/DatastoreModel.php
SurferJeffAtGoogle: Yes, for running locally.

bookshelf/src/app.php
SurferJeffAtGoogle: Leaving it false for now.  Templates contain variables that are not filled now, but will be filled in future steps.

bookshelf/src/controllers.php
SurferJeffAtGoogle: Followed the other models of never deleting files from cloud storage.

bookshelf/src/app.php
SurferJeffAtGoogle: I'll call it BOOKSHELF_DEBUG.

bookshelf/src/CloudsqlModel.php
SurferJeffAtGoogle: Exception: no such environment variable.

SurferJeffAtGoogle: I'll make sure whatever error arises displays an error about which env variable is missing.

SurferJeffAtGoogle:
Ready for another look.

tmatsuo:
LGTM

PR #5, opened 2015-12-02
Added travis config.

tmatsuo:


bshaffer:
I wouldn't mind formatting changes being in a separate PR, in the future

testing/variables.sh
```diff
@@ -0,0 +1,32 @@
...
+    GOOGLE_STORAGE_BUCKET
+    MYSQL_DSN
+    MYSQL_USER
+    MYSQL_PASSWORD
+    GOOGLE_CREDENTIALS_BASE64
+)
+
+DIRS=(
+    ${TRAVIS_BUILD_DIR}/helloworld/silex
+    ${TRAVIS_BUILD_DIR}/bookshelf
```
bshaffer: so does this mean running `testing/run_test_suite.sh` will only work on travis?

tmatsuo: Changed it to more generic TEST_BUILD_DIR

tmatsuo:
@bshaffer Yeah, it makes sense. I'll separate them next time.

tmatsuo:
Finally all the tests pass :)
@bshaffer I appreciate it if you can take a look.

testing/variables.sh
tmatsuo: Very good point, thanks!

testing/dump_credentials.php
```diff
@@ -0,0 +1,11 @@
+<?php
+/**
+ * Dumps the contents of the environment variable GOOGLE_CREDENTIALS_BASE64 to
+ * a file.
+ *
+ * To setup Travis to run on your fork, read TRAVIS.md.
```
bshaffer: we should remove this line or add a `TRAVIS.md`

tmatsuo: Good point. I added the file.

testing/install_test_dependencies.sh
```diff
@@ -0,0 +1,67 @@
...
+done
+
+# Exit when any of the necessary envvar is not set.
+if [ "${PREREQ}" = "false" ]; then
+    exit 1
+fi
+
+# Install composer
+mkdir -p ${HOME}/bin
+curl -sS https://getcomposer.org/installer | php -- --install-dir=${HOME}/bin --filename=composer
```
bshaffer: awesome, I didn't know you could send those arguments in! Those are handy.

testing/variables.sh
```diff
@@ -0,0 +1,33 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Defines some variables
+
+VARS=(
+    CLOUDSDK_ACTIVE_CONFIG_NAME
```
bshaffer: Where is this env var being used? Do we still need it?

tmatsuo: This is used by the gcloud command. It's safe to use for isolating the gcloud configs.

tmatsuo: Oh, and it's not used for now, but I will use it when I add the e2e test.

bshaffer:
LGTM

tmatsuo:
Thx!

PR #10, opened 2015-12-14
Use php-coveralls phar file from v0.7.0 release.

tmatsuo:
Removed the top level composer.json and started to use the phar file from php-coveralls 0.7.0 release.

A phar file contains all the dependencies, so no need to have composer.json anymore. Tools like php-coveralls often depend on legacy libs, which may conflict with dependencies of modern applications, so it is better to use the phar file.


tmatsuo:
Oh, wait let me add PHP 7.0 env.

tmatsuo:
Now it's all green :) @bshaffer Can you take a look?

tmatsuo:
@SurferJeffAtGoogle Alright, I added a comment in the script, also updated the PR description.

SurferJeffAtGoogle:
LGTM.

tmatsuo:
@SurferJeffAtGoogle thanks, I will merge this after squashing.

PR #11, opened 2015-12-15
Added files for MVMs deployment.

tmatsuo:


bookshelf/app.yaml
```diff
@@ -0,0 +1,12 @@
+runtime: custom
+vm: true
+
+manual_scaling:
```
jonparrott: Why this? We've removed it from the node.js and python ones intentionally.

.gitignore
```diff
@@ -5,3 +5,4 @@
 *~
 *.iml
 composer.phar
+app-prod.yaml
```
JustinBeckwith: is this storing secrets I guess?

bookshelf/app.yaml
JustinBeckwith: +1 lets get rid of this

tmatsuo: `manual_scaling`? I can remove this. Are we OK for getting-started users to have 2 instances from the start and potentially pay some unnecessary money? I prefer having a `scaling` section in the guide to show how to remove this setting if necessary?

.gitignore
tmatsuo: Yes

bookshelf/php.ini
```diff
@@ -0,0 +1 @@
+suhosin.executor.func.blacklist="escapeshellarg, escapeshellcmd, exec, highlight_file, lchgrp, lchown, link, symlink, passthru, pclose, popen, proc_close, prog_get_status, proc_nice, proc_open, proc_terminate, shell_exec, show_source, system, gc_collect_cycles, gc_enable, gc_disable, gc_enabled, getmypid, getmyuid, getmygid, getrusage, getmyinode, get_current_user, libxml_disable_entity_loader, parse_str, phpinfo, phpversion"
```
JustinBeckwith: why do we need this? In a perfect world we get away from bookshelf without needing to override php.ini or nginx.conf.

bookshelf/app.yaml
JustinBeckwith: Most new users will be using free trial money.  I fear people will start with this config, and not realize they're running single instance with a poor uptime.  I'd rather start them off with a load balanced 2 instance setup, and let them scale down if they want to.

bookshelf/php.ini
tmatsuo: Because the client library is using `php_uname()` which is disabled as a dangerous function. I agree that the current way of enabling this again is terrible though.

Re: `nginx.conf`
Having custom web server config for php apps is very common and I don't see it's an issue.

bookshelf/app.yaml
jonparrott: For Python my mantra was "show the simplest thing that works".

tmatsuo: I think `scaling` topic is important and worthwhile to explain in the guide. If we have such a section anyways, it's not an issue.

tmatsuo: I also see the downside of having 2 instances. When you have trouble in your app, how do you determine which server to ssh into? I still think we should start from 1 instance and change the scaling settings if necessary.

bookshelf/php.ini
tmatsuo: php_uname() itself is not dangerous, but it's often a first step for intruders.

Personally I'm fine with removing `php_uname()` from the blacklist by default. @sjlangley What do you think?

sjlangley: commonly used to formulate attacks - why does bookshelf need to use it?

tmatsuo: It's used by google-auth-library-php:
https://github.com/google/google-auth-library-php/blob/213ca8725ec3e3c5f5ee6560fa9f277b1cde0994/src/CredentialsLoader.php#L48

JustinBeckwith:
I don't understand why this is something you'd want to disable.  If someone has the ability to run arbitrary PHP code on my server, I kind of assume I'm already in trouble.  The usage in api client seems perfectly reasonable and common.

bookshelf/app.yaml
tmatsuo: That said, I'm fine with removing them

bookshelf/php.ini
sjlangley: Submit a pull request to use PHP_OS instead?

>>> echo PHP_OS;
Linux

sjlangley:
The list originated from our security team when we built out V1 based on
best practices when it comes to PHP security. We possible I would prefer to
default to settings based on their knowledge and experience.

Perhaps they have a different viewpoint on MVMs as the chance of people
getting p0wned is x1000 compared with V1 - I don't know.

bookshelf/php.ini
tmatsuo: Ok, that will allow us not having the php.ini. I will try that.

tmatsuo: https://github.com/google/google-auth-library-php/pull/88

tmatsuo:
I'm still curious about the reasoning of blacklisting though since PHP_OS is allowed. Do you remember the reason? Was it a possibility of buffer overflow or something, or is it just to hide the underlining OS?

.gitignore
bshaffer: is there a precedent for this? I've never seen this convention. Aren't prod credentials just stored using `gcloud auth login` and `gcloud config set project`?

bookshelf/app.yaml
bshaffer: @tmatsuo I think we should remove it for simplicity and clarity, but also include a section in getting started explaining the situation. That way, the users start simple and we use it as a teaching tool:

# Manual Scaling
By default managed VMs will deploy to 2 instances behind a load balancer. To override this, which can be preferable for this free trial because it reduces the cost of running this application, add the following to `app.yaml`:

```yaml
manual_scaling:
  instances: 1
```

.gitignore
tmatsuo: No precedent, but I think it's better to say:
```
Copy the app.yaml to app-prod.yaml and edit it.
```

Rather than:
```
Edit the app.yaml
```
Because with the latter approach, you can accidentally commit your secret to your git clone.

I can move the `app.yaml` to `app.yaml.template` and do the opposite like:

```
Copy app.yaml.template to app.yaml and edit it.
```

What do you think?

bshaffer: I like `app.yaml.template` (or `app.yaml.dist` is common in PHP), but what secrets do you expect to have in `app.yaml`?  DB passwords?

tmatsuo: Yup, also MYSQL_DSN contains the IP address of the server.

bshaffer: :+1: let's do `app.yaml.dist` then

tmatsuo: Done

bookshelf/php.ini
tmatsuo: That PR submitted. Let's not forget to stop using dev-master:
https://github.com/GoogleCloudPlatform/getting-started-php/issues/13

.gitignore
tmatsuo: Other possible idea. We can have `app.yaml` in the repo, also have `app.yaml` in `.gitignore`, so that it is very unlikely someone submit modified `app.yaml`. When we add the file, we can always use `-f`option.

How does it sound?

tmatsuo:
> I'm still curious about the reasoning of blacklisting though since PHP_OS is allowed. Do you remember the reason? Was it a possibility of buffer overflow or something, or is it just to hide the underlining OS?

I looked at the source code of php_uname which is written in C and basically a simple uname's wrapper. It will give you information including the hostname, which I think should be forbidden. Anyways, I agree anyone should use `PHP_OS` if possible, and I'm fine with the current blacklist.

sjlangley:
So what we can do is add a build step to change the result of uname to something harmless and then unblock the function. We do something like this internally

  # Use a generic nodename in the uname value
  local NODENAME=$(uname -n)
  sed -i "s/$NODENAME/unknown/" $BUILD_DIR/main/php_config.h

.gitignore
tmatsuo: Any objection?

bookshelf/app.yaml.dist
```diff
@@ -0,0 +1,12 @@
+runtime: custom
+vm: true
+
+manual_scaling:
```
JustinBeckwith: lets drop the manual_scaling directive

tmatsuo: Also I'm going to rename `app.yaml.dist` to `app.yaml` which is simpler. As long as we have `app.yaml` in the top level `.gitignore`, it won't be a problem.

tmatsuo: Well, wait. On the second thought, it is much easier for us to keep it `app.yaml.dist`, especially during our development. I will keep it for now.

tmatsuo:
Ok, I squashed the commits. I'll wait for LGTM from one of you guys.

JustinBeckwith:
LGTM

PR #12, opened 2015-12-15
remove manual scaling config

JustinBeckwith:


tmatsuo:
BTW, the test doesn't work for PR from other repo right now. Maybe I should skip the test in that case.

tmatsuo:
Ok, I'm fine with this

JustinBeckwith:
What would it take to get tests to run on forks?  @jmdobry have you figured this out in node land?

tmatsuo:
See TRAVIS.md for running tests on your project (you can run push hook). There is no solution for achieving both; running test suite for PR from other repo, and keeping the secrets.

PR #14, opened 2015-12-18
Use google/auth ~0.5.

### LINKED ISSUE #13: Stop using dev-master for google/auth

tmatsuo:
Fixed #13

### DIFF
--- a/bookshelf/composer.json
+++ b/bookshelf/composer.json
@@ -6,7 +6,7 @@
   },
   "require": {
     "google/apiclient": "v2.0.0-RC2",
-    "google/auth": "dev-master as 0.4.x-dev",
+    "google/auth": "~0.5",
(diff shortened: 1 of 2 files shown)


bshaffer:
Oh sorry, I merged in other changes as well that only work for [`google/apiclient: dev-add-support-for-guzzle-6`](https://packagist.org/packages/google/apiclient#dev-add-support-for-guzzle-6). So until *these* changes are merged into master and we update `google/apiclient`, this is going to break.

*ALSO* we should not define `google/auth` here, since it should be pulled as a dependency of `google/apiclient` only.

So yeah, I think we should close this, and for the time being check out the specific SHA of the change you need:

```json
"google/auth": "dev-master#0eaf83e as 0.4.x-dev"
```

tmatsuo:
Alrighty

PR #16, opened 2015-12-30
cleans up classes, tests, and adds Datastore

bshaffer:


bookshelf/src/DataModel/Datastore.php
```diff
@@ -0,0 +1,246 @@
...
+    ];
+
+    public function __construct($datastoreDatasetId)
+    {
+        $this->datasetId = $datastoreDatasetId;
+        $client = new \Google_Client();
+        $client->setScopes([
+            Google_Service_Datastore::CLOUD_PLATFORM,
+            Google_Service_Datastore::DATASTORE,
+            Google_Service_Datastore::USERINFO_EMAIL,
```
tmatsuo: I think we don't need these scopes any more, but I may be wrong.

bookshelf/src/DataModel/Datastore.php
```diff
@@ -0,0 +1,246 @@
...
+        ]);
+        $client->useApplicationDefaultCredentials();
+        $this->datastore = new \Google_Service_Datastore($client);
+    }
+
+    public function listBooks($limit = 10, $cursor = null)
```
tmatsuo: `10` here is a magic number. How about to make it a constant?

tmatsuo: Or even making it a URL param?

bookshelf/src/DataModel/Datastore.php
```diff
@@ -0,0 +1,246 @@
...
+        // return the number of updated rows
+        return 1;
+    }
+
+    public function delete($id)
+    {
+        $key = $this->createKey($id);
+
+        $request = new \Google_Service_Datastore_CommitRequest([
+            'mode' => 'NON_TRANSACTIONAL',
```
tmatsuo: Why not use transaction for delete and update which will prevent race?

bookshelf/src/DataModel/Datastore.php
```diff
@@ -0,0 +1,246 @@
...
+
+        $uniqueId = $this->datastore->datasets->allocateIds($this->datasetId, $keyRequest);
+
+        return $uniqueId->getKeys()[0];
+    }
+
+    private function verifyBook($book)
+    {
+        if ($invalid = array_diff_key($book, $this->columns)) {
+            throw new \Exception(sprintf(
```
tmatsuo: Question. Is `\Exception` an appropriate exception?

bookshelf/src/DataModel/Datastore.php
bshaffer: seems a little over-architecty to me

bookshelf/src/DataModel/Datastore.php
bshaffer: well that would be cool! I'll test it out.

bookshelf/src/DataModel/Datastore.php
bshaffer: Yes but it *would* be better to be more specific. In this case, `\InvalidArgumentException` would be more precise.

bookshelf/src/DataModel/Datastore.php
bshaffer: I agree. I copied this over from my datastore sample, and I don't remember adding it. I'll change it and test.

bshaffer: Ahh, here's the reason. If you want to use transactional commits, you have to wrap everything in a transaction request, which is more complicated and not really necessary for this example (we're only making one request at a time, there's no benefit in wrapping it in a transaction). Otherwise, you get this error:

```json
"domain": "global",
"reason": "INVALID_ARGUMENT",
"message": "transactional commit requires a transaction"
```

So I think we should keep it as-is

tmatsuo: Ok. I hope it will be much easier when gcloud-php is out. Can you explain it as a comment or add a TODO?

bookshelf/src/DataModel/Datastore.php
bshaffer: No, it throws an `invalid_scope` scope or a `401` when I remove all/some of them (respectively)

bookshelf/src/DataModel/Datastore.php
tmatsuo: OK.

bookshelf/src/DataModel/Datastore.php
tmatsuo: Sorry. The datastore v3 won't need USERINFO_EMAIL at least. Can you add a TODO to remove USERINFO_EMAIL when v3 is out?

bookshelf/src/DataModel/Datastore.php
bshaffer: yep! Added comments for all of these

bookshelf/src/DataModel/Datastore.php
bshaffer: Done!

tmatsuo:
LGTM

PR #18, opened 2016-01-06
add auth step

bshaffer:


bshaffer:
@tmatsuo I believe [this test](https://travis-ci.org/GoogleCloudPlatform/getting-started-php/jobs/100496636#L698) failed because of a race condition (due to parallel travis builds).  We will want to figure out a long term solution for this.

bshaffer:
cc @SurferJeffAtGoogle

bookshelf/src/controllers.php
```diff
@@ -139,3 +140,48 @@ function ($id) use ($app) {
...
+
+$app->get('/login', function () use ($app) {
+    /** @var Google_Client $client */
+    $client = $app['google_client'];
+
+    $scopes = [ \Google_Service_Oauth2::USERINFO_PROFILE ];
+    $authUrl = $client->createAuthUrl($scopes);
+
+    return $app->redirect($authUrl);
+})->bind('login');
```
tmatsuo: For other handlers, we don't call `bind` methods. Do you need to call them? Isn't it better to be consistent?

bookshelf/src/User/SimpleUser.php
```diff
@@ -0,0 +1,54 @@
...
+    public $imageUrl;
+
+    public function __construct($id = null, $name = null, $imageUrl = null)
+    {
+        $this->id = $id;
+        $this->name = $name;
+        $this->imageUrl = $imageUrl;
+    }
+
+    public function getLoggedIn()
```
tmatsuo: Nit: I feel more natural it to be `getLoggedInState` or `isLoggedIn`.

bookshelf/src/controllers.php
bshaffer: Yes, and I think we should name all our routes

bshaffer: the only one *required* to be named, though, is `login_callback`, because we have to generate that URL for oauth

bookshelf/src/User/SimpleUser.php
bshaffer: agreed, but this allows us to do `app.user.loggedIn` below. The way twig parses getters/setters, this is the simplest way to achieve that.

tmatsuo: Ah, implicit getter resolution. Ok, I can live with that.

bookshelf/src/controllers.php
tmatsuo: Filed #19

tmatsuo: Alright

tmatsuo:
LGTM

bookshelf/src/app.php
```diff
@@ -22,19 +22,47 @@
...
 ));
+
+// register the url generator
+$app->register(new UrlGeneratorServiceProvider);
+
+// turn debug on by default
 $app['debug'] = !in_array(
     getenv('BOOKSHELF_DEBUG'),
-    [false, 'false', '', '0', 'off', 'no']
+    ['false', '', '0', 'off', 'no']
```
SurferJeffAtGoogle: Good catch.

SurferJeffAtGoogle:
LGTM.

PR #20, opened 2016-01-08
adds logging, adds min PHP version, adds Datastore to index.php

bshaffer:


tmatsuo:
LGTM. The build is failing because it can not read the secret environment variables. How about to push a branch in this repo and create a PR from there?

bookshelf/web/index.php
```diff
@@ -18,14 +18,20 @@
...
 require __DIR__ . '/../src/controllers.php';
-$app['bookshelf.model'] = new CloudSql();
+
+// Cloud Storage
 $bucket = getenv('GOOGLE_STORAGE_BUCKET');
 $app['bookshelf.storage'] = new CloudStorage($bucket);
+// Cloud SQL
+$app['bookshelf.model'] = new CloudSql();
+// Cloud Datastore
+// $app['bookshelf.model'] = new Datastore($bucket);
```
tmatsuo: I noticed that you're passing $bucket to the Datastore model. Let's make it an env var.
Also how about to decide the data model backend with another env var?

bshaffer: SGTM on the env var, although when we split the steps out I think we should just uncomment stuff rather than use yet another env var.

bshaffer:
closing in favor of #21

PR #22, opened 2016-01-11
Added MongoDB backend.

tmatsuo:


bookshelf/php.ini
```diff
@@ -0,0 +1 @@
+extension=mongodb.so
```
JustinBeckwith: Curious - what's the downside to just including mongo/redis/mysql in our default php.ini?  I kind of have this goal where we don't need to override the php.ini for bookshelf :)  I don't know, maybe it's a good idea to include it just to show folks how it works.  Thoughts?

bookshelf/src/DataModel/MongoDb.php
```diff
@@ -0,0 +1,198 @@
...
+        if (false == ($namespace = getenv('MONGO_NAMESPACE'))) {
+            throw new \Exception(
+                'Set the environment variable MONGO_NAMESPACE to your '
+                . 'MongoDB namespace (in the "db.collection" format).');
+        }
+        $manager = new \MongoDB\Driver\Manager($dbUrl);
+        $this->db = new \MongoDB\Collection($manager, $namespace);
+    }
+
+    public function listBooks($limit = 10, $cursor = null)
```
JustinBeckwith: Maybe a little more giving with the comments through this function

bookshelf/src/DataModel/MongoDb.php
```diff
@@ -0,0 +1,198 @@
...
+            array_push($rows, $this->bookToArray($row));
+            $last_row = $row;
+        }
+        return array(
+            'books' => $rows,
+            'cursor' => $new_cursor,
+        );
+    }
+
+    public function create($book, $id = null)
```
JustinBeckwith: Do we typically do phpdoc headers for public functions?  I think that stuff gets fed into IDEs, so it's probably nice to be generous with type annotations and such for parameters

JustinBeckwith:
Aside from comment/phpdoc stuff, lgtm

tmatsuo:
Just made a cosmetic change

bookshelf/src/DataModel/MongoDb.php
tmatsuo: These public methods are implementations for the DataModelInterface and the interface file has comments on those methods.

bshaffer:
I think you just need to add the following to `.travis.yml`:

```yaml
service:
  - mongo
```

bookshelf/php.ini
tmatsuo: Small downside might be increased memory consumption. This can be discussed on `php-docker` repo, although I would not enable mongodb by default.

bshaffer: I would prefer us support the `ext-mongodb` directive in `composer.json` rather than enable the extension by default. But yeah, this is a conversation for `php-docker`

bookshelf/web/index.php
```diff
@@ -18,13 +18,20 @@
...
 
 $app = require __DIR__ . '/../src/app.php';
 require __DIR__ . '/../src/controllers.php';
-$app['bookshelf.model'] = new CloudSql();
+
+$db = getenv('BOOKSHELF_DATA_BACKEND');
+if ($db == 'mongodb') {
+    $app['bookshelf.model'] = new MongoDb();
+} else {
+    $app['bookshelf.model'] = new CloudSql();
```
bshaffer: I just merged in #21, so we should add a [CloudDatastore](https://github.com/GoogleCloudPlatform/getting-started-php/pull/21/files#diff-e59f036339c905e259516dfbeb2d75edR34) to this option as well, and throw an exception if the type is not valid (i.e. `BOOKSHELF_DATA_BACKEND=takashidb`)

bshaffer:
sorry, I created more work for you by merging in #21, but it should be easy to rebase and resolve:

```
git checkout master
git pull origin master
git checkout mongodb
git rebase master
# resolve conflicts and `git add` files after resolving
git rebase --continue
git push origin mongodb -f
```

bookshelf/src/DataModel/MongoDb.php
bshaffer: we *could* add a docblock with `@see DataModelInterface::listBooks` for those not using IDEs

bookshelf/src/DataModel/Datastore.php
```diff
@@ -21,9 +21,7 @@
 use Google_Service_Datastore;
 
 /**
- * Class DatastoreModel implements the DataModel with a Google Data Store.
- *
- * Incomplete and untested.
+ * Class Datastore implements the DataModel with a Google Data Store.
```
bshaffer: Haha nice catch :D

.travis.yml
```diff
@@ -36,6 +36,7 @@ env:
 
 install:
 - testing/install_test_dependencies.sh
+- echo "extension = mongodb.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
```
bshaffer: If you use the `services` directive, it will automatically activate the extension in `php.ini`, so this line is not necessary

tmatsuo:
```
service:
  - mongodb
```
didn't work for me.

bookshelf/src/DataModel/MongoDb.php
tmatsuo: will do

.travis.yml
bshaffer: sorry, it's `services` (plural), like [here](https://github.com/bshaffer/oauth2-server-php/blob/develop/.travis.yml#L19)

tmatsuo:
Now it's green with the following lines in `.travis.yml`

```
- pecl install mongodb && /bin/true
- echo "extension = mongodb.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
```

I will also try `services`

tmatsuo:
``
services:
    - mongodb
``
Doesn't work for me.

https://travis-ci.org/GoogleCloudPlatform/getting-started-php/jobs/101710071

I'll use pecl.

tmatsuo:
Ok, now the build is passing. 
@bshaffer Can you take another look?

bookshelf/web/index.php
```diff
@@ -29,10 +30,22 @@
...
+}
+if ($db == 'mongodb') {
+    $app['bookshelf.model'] = new MongoDb();
+} elseif ($db == 'datastore') {
+    $datasetId = getenv('GOOGLE_DATASET_ID');
+    $app['bookshelf.model'] = new Datastore($datasetId);
+} elseif ($db == 'cloudsql') {
+    $app['bookshelf.model'] = new CloudSql();
+} else {
+    throw \Exception("Invalid BOOKSHELF_DATA_BACKEND given: $db. "
```
bshaffer: the `\` is not necessary, since this file does not have a namespace

bshaffer:
LGTM!

tmatsuo:
@bshaffer Thanks. I squashed the commits and I'll merge once it's green.

PR #23, opened 2016-01-12
bug fixes for datastore

bshaffer:


bshaffer:
argh... why do I even have a fork??

PR #24, opened 2016-01-12
bug fixes for datastore

bshaffer:


bookshelf/web/index.php
```diff
@@ -24,8 +24,14 @@
 
 $app = require __DIR__ . '/../src/app.php';
 require __DIR__ . '/../src/controllers.php';
-$app['bookshelf.model'] = new CloudSql();
+
```
tmatsuo: I feel like this diff shows a diff against older file contents. I don't know why.

bshaffer: hmm yeah that isn't right. I'll look into it.

bshaffer: ok, fixed

bookshelf/src/app.php
```diff
@@ -41,7 +41,8 @@
 // turn debug on by default
 $app['debug'] = !in_array(
     getenv('BOOKSHELF_DEBUG'),
-    ['false', '', '0', 'off', 'no']
+    ['false', '', '0', 'off', 'no'],
+    true
```
tmatsuo: In my opinion, the value of `$app['debug']` must be by default `false`. `debug=true` is a dangerous configuration on prod because it may reveal your secrets (db password etc).

bshaffer: Yeah I agree, it'd be better to guide developers with our logging/debugging tools.

tmatsuo:
LGTM

PR #25, opened 2016-01-14
RFC - adds GCE

bshaffer:


bookshelf/gce/deploy.sh
```diff
@@ -0,0 +1,151 @@
...
+# We'll create these resources in reverse order:
+# service, health check, backend service, url map, proxy.
+
+# Create a health check
+# The load balancer will use this check to keep track of which instances to send traffic to.
+# Note that health checks will not cause the load balancer to shutdown any instances.
+
+# [START create_health_check]
+gcloud compute http-health-checks create ah-health-check \
+  --request-path /_ah/health
```
tmatsuo: Do we have /_ah/health path configured in our app? If we don't and the app returns HTTP 404, is it still considered to be healthy?

bookshelf/gce/deploy.sh
```diff
@@ -0,0 +1,151 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#! /bin/bash
+
+set -ex
+
+ZONE=us-central1-f
```
tmatsuo: Is it possible to make it configurable with an environment variable?

bookshelf/gce/nginx/fastcgi_params
```diff
@@ -0,0 +1,41 @@
...
+fastcgi_param   PATH_TRANSLATED         $document_root$fastcgi_path_info;
+fastcgi_param   REQUEST_URI             $request_uri;
+fastcgi_param   DOCUMENT_URI            $document_uri;
+fastcgi_param   DOCUMENT_ROOT           $document_root;
+fastcgi_param   SERVER_PROTOCOL         $server_protocol;
+
+fastcgi_param   GATEWAY_INTERFACE       CGI/1.1;
+fastcgi_param   SERVER_SOFTWARE         nginx/$nginx_version;
+
+fastcgi_param   REMOTE_ADDR             $http_x_real_ip;
```
tmatsuo: Is REMOTE_ADDR set as you desire by this configuration?

bookshelf/gce/startup-script.sh
```diff
@@ -0,0 +1,74 @@
...
+
+# [START all]
+set -e
+
+# Talk to the metadata server to get the project id
+PROJECTID=$(curl -s "http://metadata.google.internal/computeMetadata/v1/project/project-id" -H "Metadata-Flavor: Google")
+
+# Get the source code
+export HOME=/root
+git config --global credential.helper gcloud.sh
```
tmatsuo: Is `git` always available?

tmatsuo: I mean, you're running apt-get later anyways, maybe you can move the apt-get invocation up to here add installing git.

bookshelf/gce/startup-script.sh
```diff
@@ -0,0 +1,74 @@
...
+apt-get install -y php5 nginx build-essential php5-fpm # libpcre3 libpcre3-dev zlib1g-dev
+
+# Install composer
+curl -sS https://getcomposer.org/installer | \
+    /usr/bin/php -- \
+    --install-dir=/usr/local/bin \
+    --filename=composer
+
+# run composer
+composer install -d /opt/app/bookshelf --no-ansi --no-progress
```
tmatsuo: How about to add `--no-dev`?

bookshelf/gce/startup-script.sh
```diff
@@ -0,0 +1,74 @@
...
+# [END php]
+
+# [START nginx]
+# disable the default nginx configuration
+rm /etc/nginx/sites-enabled/default
+
+# enable our nginx configuration
+cat /opt/app/bookshelf/gce/nginx/bookshelf.conf > /etc/nginx/sites-available/bookshelf.conf
```
tmatsuo: Why not just `cp`?

bookshelf/gce/teardown.sh
```diff
@@ -0,0 +1,41 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#! /bin/bash
+
+set -x
+
+ZONE=us-central1-f
```
tmatsuo: ditto

tmatsuo:
Re: gzip config

I would add [the gzip config](https://github.com/GoogleCloudPlatform/php-docker/blob/master/php-nginx/gzip_params) because it will affect the network bandwidth, thus resulting higher cost, especially if you get massive load to your application. Why don't you just add it?

Or is gzip enabled by default?

bookshelf/gce/startup-script.sh
tmatsuo: We have mongodb data backend module. Can you also install mongodb extension?

bookshelf/gce/startup-script.sh
```diff
@@ -0,0 +1,74 @@
...
+  tag bookshelf
+</source>
+EOF
+service google-fluentd restart &
+# [END logging]
+
+# [START php]
+# Install PHP and dependencies from apt
+apt-get update
+apt-get install -y php5 nginx build-essential php5-fpm # libpcre3 libpcre3-dev zlib1g-dev
```
tmatsuo: It doesn't seem you are building something with the compiler. Do you need build-essential?

bookshelf/gce/startup-script.sh
tmatsuo: If you do that, you may or may not need the build-essential after all.

bookshelf/gce/deploy.sh
bshaffer: I don't know the answer to that, but we should just add a health-check route to silex.

bookshelf/gce/startup-script.sh
bshaffer: yeah, git is already installed. I could remove installing `git` again via `apt-get`, but other runtimes do it this way. I agree, it is confusing.

JustinBeckwith:
So high level question - why are we orchestrating all of this via shell scripts and gcloud commands?  Why not a deployment manager template?

bookshelf/gce/startup-script.sh
bshaffer: I saw it done [here](https://github.com/GoogleCloudPlatform/getting-started-ruby/blob/master/7-compute-engine/gce/configure.sh#L50) and decided to stay consistent

bookshelf/gce/deploy.sh
```diff
@@ -0,0 +1,151 @@
...
+
+GROUP=frontend-group
+TEMPLATE=$GROUP-tmpl
+MACHINE_TYPE=f1-micro
```
JustinBeckwith: isn't that going to run kind of bad?

bshaffer:
I believe the reason is to illustrate customizing a simple deployment on GCE. And also to stay consistent with the other getting started tutorials.

We could add a step for using a template as well, if that is preferred.

bookshelf/gce/startup-script.sh
tmatsuo: I mean, you should do:

```
apt-get update
apt-get install -y php5 nginx build-essential php5-fpm git
```

This way it won't break if the default base image changes right?

JustinBeckwith:
Another high level thing - why do we choose to do a fully scalable solution with HTTP LBs and auto scalers?  I'd imagine a lot of people that end up here are saying "I just want to run a simple PHP app without App Engine 'stuff', how do I get this running on a VM".  I think both are valuable, but is this what we do elsewhere?

bshaffer:
I agree, I don't really particularly want to put the generic `deploy.sh` and `teardown.sh` scripts in there. I'm copying what the other getting started tutorials have done. But I'd be more than happy to get rid of them, or move them to a single generic repo (since they're not language-specific)

bookshelf/gce/startup-script.sh
tmatsuo: It's not a good reason. Also you're using cp below :)

bookshelf/gce/deploy.sh
bshaffer: yeah it's possible. I copied these scripts directly from the other getting-started repos.

JustinBeckwith:
Yup.  Lets be consistent for now, and then rationalize this approach across languages next week.

bookshelf/gce/startup-script.sh
bshaffer: agreed, will change to `cp`!

bookshelf/gce/deploy.sh
bshaffer: It's the same one we are using for the rest of the tutorial. Presumably because it's cheapest.

And for the size of this app, it runs just fine :smile:

bookshelf/gce/startup-script.sh
```diff
@@ -0,0 +1,83 @@
...
+curl -sS https://getcomposer.org/installer | \
+    /usr/bin/php -- \
+    --install-dir=/usr/local/bin \
+    --filename=composer
+
+# Get the source code
+git config --global credential.helper gcloud.sh
+
+# Change branch from master if not using master
+git clone https://source.developers.google.com/p/$PROJECTID /opt/app -b add-compute-engine
```
bshaffer: **NOTE** - the `-b` option must be changed before merging!!

bshaffer:
closing as I've made quite a few changes. Resubmitting the new one momentarily.

bshaffer:
see #27

PR #26, opened 2016-01-16
refactors settings to use yaml instead of envvars

bshaffer:


tmatsuo:
It might be simpler to use the yaml file at a first glance, but I'm concerned that we will lost the capability of changing the configuration by the environment variables.

Now the travis build fails with coding syntax, but does it work with the current default yaml config file?

Also how can you change the settings for various environments like prod/staging/qa/local?

The option comes to mind first is to introduce another env variable `BOOKSHELF_YAML` deefaults tp `settings.yaml` so that you can change it to something like `settings-qa.yaml`, `settings-staging.yaml`, and `setting-local.yaml` etc to change it if necessary.

There is still a concern about travis build. We have secrets environment variables configured for the build. Those are `MYSQL_USER`, `MYSQL_PASSWORD`, and `MYSQL_DSN`, etc. If we only use the contents in the settings yaml file (regardless we add the above env var), we need to commit those secrets to the repo.

The option comes to mind is to keep using those environment variables for the travis builds, since the config is just an object, so you can still use those environment variables in the tests.

tmatsuo:
Aside from the above 2 points, the code looks great, but please address those issues before submitting :)

bshaffer:
I don't think an environmentally-aware settings file is required for the bookshelf app.Having the one file the users can modify to run locally seems sufficient, and our tutorial doesn't require the app to be run locally. How do the other bookshelf apps do it?

The test one makes sense and I will fix that right away!

tmatsuo:
The guide for other languages usually explain how to run the app locally.

For example, 
https://cloud.google.com/python/getting-started/using-cloud-sql has a section "Run the app on your local computer".

Anyways, I can live with the fixed file name for now.

bshaffer:
ok, I see. I will add the environment variable then!

bshaffer:
@tmatsuo Python has a "Run the app locally", but it uses the same `config.py` file.

tmatsuo:
Well, while I actually still prefer allowing to change the configuration based on the env variable, I can live with not having it on the getting started guide. I think we should introduce that mechanism in the code (while we won't use it in the getting started guide, it is great if we can say "Use the env variable" for people who wants to change the configurations), but this can be done with another PR.

bookshelf/test/lib/GetConfigTrait.php
```diff
@@ -0,0 +1,34 @@
...
+ * Class GetConfigTrait
+ * @package Google\Cloud\Samples\Bookshelf
+ *
+ * Use this trait to load the project configuration
+ */
+trait GetConfigTrait
+{
+    protected function getConfig()
+    {
+        return Yaml::parse(__DIR__ . '/../../config/settings.yml');
```
tmatsuo: I think the config should be obtained from the env variables for the tests?

bookshelf/test/lib/GetConfigTrait.php
```diff
@@ -0,0 +1,53 @@
...
+ * Use this trait to load the project configuration
+ */
+trait GetConfigTrait
+{
+    protected function getConfig()
+    {
+        $config = [];
+
+        // if the settings.yml file exists, copy it over
+        if (file_exists(__DIR__ . '/../../config/settings.yml')) {
```
tmatsuo: Do you really need to read the configuration file in the test and merge it with the envvars? In my opinion, you can just create an array from the env vars can't you?

bshaffer: My reasoning here was for config items that we don't need to change w/ EnvVars (i.e. `bookshelf_backend`. Also, if a user sets their local config, they can run tests against it.

I think it's better to keep it, but I'm mostly ambivalent. Is your argument here for keeping it simple?

tmatsuo: Mine is not a strong opinion, I just thought it was easier to make the tests green :)

bshaffer:
@tmatsuo fixed tests and simplified config. PTAL!

tmatsuo:
LGTM :+1:

PR #28, opened 2016-01-22
Adds e2e

tmatsuo:
Now the test should pass, currently it only checks successful deploy and status code 200 for the `/book` url, but it is ready for your review @bshaffer


bookshelf/test/lib/E2EDeploymentTrait.php
```diff
@@ -0,0 +1,101 @@
...
+        }
+        return sprintf(
+            'https://%s-dot-%s.appspot.com',
+            self::$versions[$step],
+            $projectId
+        );
+    }
+}
+E2EDeploymentTrait::$versions = array();
+E2EDeploymentTrait::$gcloud = new Gcloud();
```
bshaffer: the `$versions` can be done inside the class declaration, i.e.:

```php
public static $versions = array();
```

bookshelf/src/app.php
```diff
@@ -39,8 +39,11 @@
 // register the url generator
 $app->register(new UrlGeneratorServiceProvider);
 
+$config = getenv('BOOKSHELF_CONFIG') ?: 'settings.yml';
```
bshaffer: It would be more intuitive to make the the full path to the file, rather than just the filename

bookshelf/test/lib/E2EDeploymentTrait.php
```diff
@@ -0,0 +1,101 @@
...
+    /** @staticvar Gcloud $gcloud */
+    public static $gcloud;
+
+    private static function dumpConfig()
+    {
+        $config = self::getConfig();
+        $dumper = new Dumper();
+        $yaml = $dumper->dump($config);
+        // TODO: Use different filename
+        file_put_contents(__DIR__ . '/../../config/settings-e2e.yml', $yaml);
```
bshaffer: Keeping the settings separate seems like a good idea, although it's kinda odd to do this since we're using the same credentials in both places anyway.

bookshelf/test/e2e/BookshelfTest.php
```diff
@@ -0,0 +1,58 @@
...
+use Behat\Mink\Session;
+
+/**
+ * Class BookshelfTest
+ */
+class BookshelfTest extends \PHPUnit_Framework_TestCase
+{
+    use E2EDeploymentTrait;
+
+    const STEP = 'bookshelf';
```
bshaffer: Just a heads up, this was a placeholder until I added directory support. So I'll refactor this out once we merge it.

bookshelf/src/app.php
tmatsuo: Isn't it tricky to do that? I mean, the full path will be `/app/config/settings.yml` and of course differs in the local environment.

bookshelf/test/lib/E2EDeploymentTrait.php
tmatsuo: No, I actually use the different MONGO_COLLECTION for e2e and the deployed app. Otherwise e2e test will pollute our production deployment.

bookshelf/test/e2e/BookshelfTest.php
tmatsuo: Sure

bookshelf/test/lib/E2EDeploymentTrait.php
tmatsuo: Done

bookshelf/src/app.php
tmatsuo: nm, I agreed and changed it to use the full path.

bookshelf/test/lib/E2EDeploymentTrait.php
bshaffer: aha, sounds good.

tmatsuo:
@bshaffer Anything left?

bshaffer:
Was just waiting for tests to pass... LGTM!

tmatsuo:
Thanks, merging!

PR #29, opened 2016-01-22
Adds directories

bshaffer:


tmatsuo:
I don't think the test will pass. Now each directory has its own `.travis.yml`, but there should be one in the top level.

bshaffer:
@tmatsuo I will remove those.

bshaffer:
I suppose since each step is supposed to be taken as a standalone app, it's ok to have travis files in each directory. What do you think?

tmatsuo:
Travis doesn't look at them, so those are nothing to do with our test run. I would remove them.

bshaffer:
closing in favor of #30 so we can run tests

PR #30, opened 2016-01-25
Adds directories

bshaffer:


1-hello-world/composer.json
```diff
@@ -1,7 +1,7 @@
 {
   "require": {
-    "silex/silex": "^1.3",
-    "php": "^5.6"
+    "php": "^5.5",
```
tmatsuo: We don't currently have PHP 5.5 in the docker image.

bshaffer: That's ok! This app is able (and needs to be able) to run on `PHP 5.5`.

.travis.yml
```diff
@@ -33,11 +32,18 @@ env:
     - GOOGLE_APPLICATION_CREDENTIALS=${TRAVIS_BUILD_DIR}/credentials.json
     - CLOUDSDK_ACTIVE_CONFIG_NAME=getting-started-php
     - TEST_BUILD_DIR=${TRAVIS_BUILD_DIR}
+  matrix:
+    - STEP_NAME=1-hello-world
+    - STEP_NAME=2-structured-data
+    - STEP_NAME=3-cloud-storage
+    - STEP_NAME=4-auth
+    - STEP_NAME=5-logging
+    - STEP_NAME=6-compute-engine
```
tmatsuo: This will cause total of 18 matrices (3 php vers x 6 steps) which I'd like to avoid.

Can you keep the current PHP version based matrix and add the new subdirectories in the main phpunit.xml?

2-structured-data/composer.json
```diff
@@ -0,0 +1,22 @@
...
+  "autoload": {
+    "psr-4": {
+      "Google\\Cloud\\Samples\\Bookshelf\\": ["src", "test/lib"]
+    }
+  },
+  "require": {
+    "php": "^5.5",
+    "silex/silex": "^1.3",
+    "twig/twig": "^1.22",
+    "mongodb/mongodb": "^1.0.0@beta",
```
tmatsuo: Can you use ^1.0.0 ?

tmatsuo:
You need to update the .coveralls.yml file.
Maybe we can get coveralls off from the script section, maybe into after_script section?

bshaffer:
@tmatsuo PTAL

tmatsuo:
LGTM

I think you intentionally have the test code in the sub directories with some code duplication. I actually prefer this approach over over-optimizations.

bshaffer:
Yes, the reason is so that you can run the full test suite for each step. The code duplication is a little unfortunate, but I believe it's well worth it.

tmatsuo:
I don't think we should do this right now, but maybe we can release our testing library (test/lib/*) someday. What do you think?

bshaffer:
@tmatsuo you mean make it a stand-alone repo for deploying e2e tests to GCP?

PR #32, opened 2016-01-28
Dependency fixes.

tmatsuo:


bshaffer:
The composer.lock files also need updated here.

tmatsuo:
I see, are you going to create another PR?

bshaffer:
yep: #36

PR #52, opened 2016-02-04
fixes compute engine for directories

bshaffer:


bshaffer:
These changes have been published! View these changes [here](https://github.com/GoogleCloudPlatform/getting-started-php/compare/master...fix-compute-engine-published).

tmatsuo:
Can you also have the fluentd config as a file in this repo instead of using inlined string?

tmatsuo:
LGTM

PR #65, opened 2016-02-06
Enables e2e test when RUN_E2E_TEST envvar is set.

tmatsuo:
This shouldn't trigger e2e tests on travis unless explicitly set that envvar. I will configure jenkins jobs after merging this.


bshaffer:
LGTM'd!

PR #72, opened 2016-02-09
Added PHP 5.6, and composer clean-ups

tmatsuo:
I wanted to add PHP 5.6 explicitly. Also I did composer cleanups (alphabetical order).

### DIFF
--- a/_tuts/1-hello-world.diff
+++ b/_tuts/1-hello-world.diff
@@ -29,7 +29,7 @@ index 0000000..67656f0
 \ No newline at end of file
 diff --git a/composer.json b/composer.json
 new file mode 100644
-index 0000000..66aef78
+index 0000000..6ff2189
 --- /dev/null
 +++ b/composer.json
 @@ -0,0 +1,18 @@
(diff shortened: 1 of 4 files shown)


bshaffer:
These changes have been published! View these changes [here](https://github.com/GoogleCloudPlatform/getting-started-php/compare/master...composer-cleanup-published).

bshaffer:
So, `^5.5` actually includes PHP 5.6, which is why I left it that way. I don't mind this though, because it's explicit.

bshaffer:
see the [composer documentation](https://getcomposer.org/doc/articles/versions.md#caret) for more info

bshaffer:
If you want to be explicit, the correct composer use would be `~5.5.0|~5.6.0|~7.0.0`, which essentially means "anything between `5.5.0` and < `5.6.0`, anything between `5.6.0 and < 5.7.0`, and anything between `7.0.0 and < 7.1.0`"

tmatsuo:
Alright, thanks, done. PTAL @bshaffer

bshaffer:
LGTM!

PR #78, opened 2016-02-10
update google/api-client to RC5

bshaffer:


bshaffer:
These changes have been published! View these changes [here](https://github.com/GoogleCloudPlatform/getting-started-php/compare/master...update-google-api-php-client-RC5-published).

tmatsuo:
On travis, I saw in the `2-structured-data`, it still downloads RC4.

bshaffer:
this is true... because I didn't think we used the library in step 2... :( 

good catch

tmatsuo:
Only concern is that, the `tutorial_steps` branch was rolled back to sometime in the past (see https://github.com/GoogleCloudPlatform/getting-started-php/pull/80) so it will conflict if we merge this first.

If you don't mind, I will do `composer update` on my local branch and add an equivalent change there, or I will create another PR. What do you think?

tmatsuo:
NM, LGTM

PR #80, opened 2016-02-11
Phpseclib workaround

tmatsuo:


bshaffer:
These changes have been published! View these changes [here](https://github.com/GoogleCloudPlatform/getting-started-php/compare/master...phpseclib-workaround-published).

tmatsuo:
Seems like the current state of the branch `tutorial_steps` was roll backed to sometime in the past.

bshaffer:
We are going to add this to the php client library instead

PR #82, opened 2016-02-11
(Really) allow pecl install mongodb to fail.

tmatsuo:


bshaffer:
if the mongo install fails, the unit tests will fail. Why do we want to add this?

tmatsuo:
@bshaffer It fails when the underlining travis image already has the mongodb extension installed. Actually travis image was started to have mongodb.so in their image at some point. I want to be resilient for such changes.

bshaffer:
ok, LGTM

tmatsuo:
Thanks!

PR #87, opened 2016-02-12
[PUBLISHED] 86 - Added e2e tests for every backend

tmatsuo:


bshaffer:
LGTM!!  However, @broady and I have an app for scheduling tests. We can use the environment variable `SCHEDULED_BUILD=true` and run E2E tests if it's set.

tmatsuo:
If an envvar `RUN_E2E_TEST` is set, it should run the e2e tests.

PR #92, opened 2016-02-12
[PUBLISHED] 91 - Adding e2e tests for every backend.

tmatsuo:
Trying to run the e2e test on travis.


tmatsuo:
I temporary set the `RUN_E2E_TEST` envvar, now PHP5.5 build is running the e2e test.

tmatsuo:
The previous tests failed because of transient datastore errors. Let's see if it passes.

tmatsuo:
The weird things are happening on travis. The build is suddenly gone and restarted, which results in stale builder vms and the following tests fails. At a glance, the tests are theoretically correct though.

tmatsuo:
Alright, the tests don't seem to finish at all. It suddenly restarted many times. I think this is the reason why we should run those slow and heavy tests on our infrastructure.

I will wait for few more days to see how it goes, then will switch turning off the e2e test on travis.

tmatsuo:
I'm seeing weird sporadic memory errors like:

```
Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 64 bytes) in phar:///usr/local/bin/composer/src/Composer/DependencyResolver/RuleSetGenerator.php on line 126
Check https://getcomposer.org/doc/articles/troubleshooting.md#memory-limit-errors for more info on how to handle out of memory errors.
```

tmatsuo:
Oh now I see the reason:

```
The job exceeded the maxmimum time limit for jobs, and has been terminated.
```

tmatsuo:
Alright, it's now realistic to run all the e2e test on travis. I'm going to unset the `RUN_E2E_TEST` and restart the travis build.

tmatsuo:
Some functional tests are flaky (e.g. datastore service returns 503 error). We should find a way for retrying/marking them as flaky.

2-structured-data/test/e2e/BookshelfTestBase.php
```diff
@@ -21,15 +21,20 @@
...
- * Class BookshelfTest
+ * Class BookshelfTestBase
  */
-class BookshelfTest extends \PHPUnit_Framework_TestCase
+abstract class BookshelfTestBase extends \PHPUnit_Framework_TestCase
 {
     protected static $step;
     use E2EDeploymentTrait;
 
+    protected static function setBackEnd()
```
bshaffer: This could be `abstract` to require each subclass to set it

2-structured-data/test/e2e/BookshelfTestBase.php
```diff
@@ -21,15 +21,20 @@
 use Behat\Mink\Session;
 
 /**
- * Class BookshelfTest
+ * Class BookshelfTestBase
  */
-class BookshelfTest extends \PHPUnit_Framework_TestCase
+abstract class BookshelfTestBase extends \PHPUnit_Framework_TestCase
```
bshaffer: This would more appropriately live in `test/lib/`. Also, I'd prefer `BaseTest` for consistency with [google-api-php-client](https://github.com/google/google-api-php-client/blob/master/tests/BaseTest.php)

2-structured-data/test/e2e/CloudSqlTest.php
```diff
@@ -0,0 +1,41 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Cloud\Samples\Bookshelf;
+
+require_once('BookshelfTestBase.php');
```
bshaffer: If the base class is moved to `test/lib`, it will be picked up by the autoloader (see [composer.json](https://github.com/GoogleCloudPlatform/getting-started-php/blob/e2e-published/2-structured-data/composer.json#L4)), and then we can remove this line.

2-structured-data/test/e2e/BookshelfTestBase.php
bshaffer: Actually, should it be named `E2eTest`?

2-structured-data/test/e2e/BookshelfTestBase.php
tmatsuo: I initially did that, but this will raise a `strict` warning.

2-structured-data/test/e2e/BookshelfTestBase.php
tmatsuo: Alright, I can move this to `test/lib` and rename it to E2eTest

bshaffer:
One suggestion for getting around the travis issue - in the `install` part of our travis config, run a script to deploy to `gcloud` for all the steps in parallel. Travis has [rabbitmq](https://docs.travis-ci.com/user/database-setup/#RabbitMQ) that could be used for this. Then block on all the jobs finishing (using `$channel->wait()` as defined [here](https://www.rabbitmq.com/tutorials/tutorial-one-php.html)) and run once they've completed. This would cut the time down significantly.

tmatsuo:
@bshaffer It can solve the deadline issue, but it doesn't solve mysterious restarts. We should be able to investigate issues further when something wrong happens. Now we can just retry the job.

tmatsuo:
@bshaffer I renamed and moved the base class. PTAL

bshaffer:
LGTM

tmatsuo:
Thanks!

PR #95, opened 2016-02-17
Dynamically create a firewall rule for test.

### LINKED ISSUE #94: Dynamically create a firewall rule for tests

tmatsuo:
fixes #94

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,9 @@ env:
     - PATH=${HOME}/bin:${GCLOUD_DIR}/google-cloud-sdk/bin:${PATH}
     - GOOGLE_APPLICATION_CREDENTIALS=${TRAVIS_BUILD_DIR}/credentials.json
     - CLOUDSDK_ACTIVE_CONFIG_NAME=getting-started-php
+    - CLOUDSDK_CORE_DISABLE_PROMPTS=1
     - TEST_BUILD_DIR=${TRAVIS_BUILD_DIR}
+    - FIREWALL_PREFIX=travis-test
 
 install:
 - testing/install_test_dependencies.sh
@@ -44,3 +46,6 @@ install:
 
 script:
 - testing/run_test_suite.sh
+
+after_script:
+- testing/test_cleanup.sh
--- a/testing/install_test_dependencies.sh
+++ b/testing/install_test_dependencies.sh
@@ -92,3 +92,16 @@ else
     gcloud auth activate-service-account --key-file \
         "${GOOGLE_APPLICATION_CREDENTIALS}"
 fi
+
+TIMESTAMP=$(date +%s)
+export FIREWALL_NAME="${FIREWALL_PREFIX}-${TIMESTAMP}"
+
+gcloud compute firewall-rules delete ${FIREWALL_NAME} || /bin/true
+sleep 10
+# Create a firewall rule for mongodb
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -44,3 +45,6 @@ install:
 
 script:
 - testing/run_test_suite.sh
+
+after_script:
```
bshaffer: I think it'd be better to make the script name `test_cleanup.sh` or something along those lines, to match with the generic `install_test_dependencies.sh`. That way, if we have additional cleanup tasks/scripts, we can add them there.

bshaffer:
one minor suggestion, otherwise LGTM

.travis.yml
tmatsuo: Makes sense, will do

tmatsuo:
Thanks! I will merge once the test passes.

tmatsuo:
Oops it hangs with:

```
gcloud compute firewall-rules delete ${FIREWALL_NAME} || /bin/true
The following firewalls will be deleted:
 - [travis-test]
Do you want to continue (Y/n)? 
```

I should add `CLOUDSDK_CORE_DISABLE_PROMPTS=1` envvar. I will update the PR.

bshaffer:
good call

tmatsuo:
The test failed. Very weird.

```
Created [https://www.googleapis.com/compute/v1/projects/getting-started-php/global/firewalls/travis-test].
NAME        NETWORK SRC_RANGES        RULES     SRC_TAGS TARGET_TAGS
travis-test default 104.197.119.59/32 tcp:27017          mongodb
```

```
There was 1 error:
1) Google\Cloud\Samples\Bookshelf\DataModel\MongoDbTest::testDataModel
MongoDB\Driver\Exception\RuntimeException: No suitable servers found (`serverselectiontryonce` set): [Failed connecting to '104.197.14.71:27017': Connection timed out]
```

```
$ testing/test_cleanup.sh
ERROR: (gcloud.compute.firewall-rules.delete) Some requests did not succeed:
 - The resource 'projects/getting-started-php/global/firewalls/travis-test' was not found
```

It is confusing, possible interpretation is that the creation somehow failed, but gcloud reported success.

tmatsuo:
I found this

```
gcloud compute firewall-rules delete ${FIREWALL_NAME} || /bin/true
Deleted [https://www.googleapis.com/compute/v1/projects/getting-started-php/global/firewalls/travis-test].
```

Just before the creation call above. Another interpretation is that those 2 calls were conflicting each other, or executed in the reverse order. I added a sleep call after the deletion.

tmatsuo:
Just a record, failure log is available at:
https://travis-ci.org/GoogleCloudPlatform/getting-started-php/jobs/109977272

tmatsuo:
I came up with more possible hypothesis. It is possible that there were some stale build which execute the deletion code. I think this is true because even when there is no travis build running (at least travis UI said so), I saw a firewall `travis-test` existed on the developers console.

I will add timestamp to the firewall name.

tmatsuo:
The hypothesis in https://github.com/GoogleCloudPlatform/getting-started-php/pull/95#issuecomment-185447635 seems correct. The firewall `travis-test` magically disappeared now.

bshaffer:
hmm, very interesting. We have them only allowing one build at a time, so I wonder how this could have happened. In any case, nice work.

tmatsuo:
Alright, merging

PR #97, opened 2016-02-17
[PUBLISHED] 96 - updates apiclient to dev tag

bshaffer:


tmatsuo:
LGTM, filed an issue to use stable at #98

bshaffer:
Good call, ty

tmatsuo:
The build will just fail because there is no firewall rule, so I stopped the build.

tmatsuo:
After I merge #95, I will manually add CLOUDSDK_CORE_DISABLE_PROMPTS=1 envvar and restart the build.

tmatsuo:
I restarted the build

tmatsuo:
Alright, I think we can merge this :)

PR #102, opened 2016-02-18
remove gitignore and README from start step

bshaffer:


bshaffer:
These changes have been published! View these changes [here](https://github.com/GoogleCloudPlatform/getting-started-php/compare/master...remove-gitignore-and-readme-from-start-step-published).

bshaffer:
This essentially just makes the starting directory clean. This is a precursor to using `master` as the starting step (changes I am making to the tutorial tooling)

tmatsuo:
LGTM, but why the webhook publishing doesn't work for it?

bshaffer:
I deleted it (sorry) because it doesn't have any changes for the published branch, since I basically just reordered the diff.

PR #109, opened 2016-02-18
[PUBLISHED] 108 - Added an instance_tag.

tmatsuo:


bshaffer:
Why is this tagged specifically for mongo? Doesn't that apply to the whole app and not just the mongo instance?

tmatsuo:
Good point, I've just changed the tag name to bookshelf-e2e

tmatsuo:
OK to merge?

bshaffer:
LGTM!

PR #112, opened 2016-02-18
[PUBLISHED] 111 - add retries for datastore

bshaffer:


2-structured-data/src/DataModel/Datastore.php
```diff
@@ -39,7 +39,10 @@ class Datastore implements DataModelInterface
     public function __construct($datastoreDatasetId)
     {
         $this->datasetId = $datastoreDatasetId;
-        $client = new \Google_Client();
+        // Datastore API has intermittent failures, so we set the
+        // Google Client to retry in the event of a 503 Backend Error
+        $retryConfig = [ 'retries' => 2 ];
```
tmatsuo: I would retry 3 times, but I can live with this

tmatsuo:
LGTM

PR #114, opened 2016-02-19
[PUBLISHED] 113 - Override mongo_url setting for e2e test

tmatsuo:
This is needed because we use VMs' instnace tag for the firewall rule and when we do that, we need to use the mongodb server's internal IP address, while we still need to use external IP address for the functional tests.


bshaffer:
LGTM!

PR #116, opened 2016-02-20
adds _tuts steps to master

bshaffer:


tmatsuo:
LGTM

tmatsuo:
Basically I'm all for the change. I'm wondering though, why this _tuts dir doesn't contain the same files as the current tutorial_steps? Will you add the diffs later?

bshaffer:
I coped the files over from tutorial_steps, so it is the same! Are you
seeing something differently?

tmatsuo:
Oh I see. LGTM

Issue #107, opened 2016-02-18
MongoDB warning on Travis


bshaffer:
on Travis (see [this build](https://travis-ci.org/GoogleCloudPlatform/getting-started-php/jobs/110051836#L2401)):

```
PHP Warning:  Module 'mongodb' already loaded in Unknown on line 0
Warning: Module 'mongodb' already loaded in Unknown on line 0
```

This may be because the `mongodb` extension is already compiled into PHP, and we are also installing it in [.travis.yml](https://github.com/GoogleCloudPlatform/getting-started-php/blob/master/.travis.yml#L44) via `pecl`

tmatsuo:
This is fine in my opinion. Travis doesn't guarantee the existence of the mongodb extension and I want to be resilient to their changes.

bshaffer:
While I don't typically agree with fixing a problem before it exists, I've added a conditional so we can at least get rid of the warnings being thrown in our tests.

PR #117, opened 2016-02-20
only install mongodb extension if we need to

bshaffer:
addresses #107

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,8 +41,10 @@ env:
 
 install:
 - testing/install_test_dependencies.sh
-- pecl install mongodb || /bin/true
-- phpenv config-add testing/mongodb.ini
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -41,8 +41,10 @@ env:
 
 install:
 - testing/install_test_dependencies.sh
-- pecl install mongodb || /bin/true
-- phpenv config-add testing/mongodb.ini
+- if [ ! `php -r 'echo extension_loaded("mongodb");'` ]; then
+  pecl install mongodb || /bin/true;
```
tmatsuo: Thanks, in this case, if we failed to install mongodb, the test can fail.
Can you remove `|| /bin/true` part?

tmatsuo: Ah, but there is still a case mongodb extension is available as a shared module.
Please just ignore the comment above.

LGTM

PR #118, opened 2016-02-24
Cleaner way of handling BOOKSHELF_DEBUG envvar

tmatsuo:


bshaffer:
LGTM!

tmatsuo:
Thanks, merging

_tuts/steps.json
```diff
@@ -37,4 +37,4 @@
         }
     ],
     "sha": false
-}
+}
```
bshaffer: I need to fix this :-/

PR #119, opened 2016-02-25
Turn on debug locally

tmatsuo:


bshaffer:
Hmm.. I don't really like this because they may or may not be using `localhost`. I.e. they could be using `127.0.0.1` or have updated `/etc/hosts`. Both of these are pretty common. 

This is [how symfony does it](https://github.com/symfony/symfony-standard/blob/master/web/app_dev.php#L13-L19):
```php
if (isset($_SERVER['HTTP_CLIENT_IP'])
    || isset($_SERVER['HTTP_X_FORWARDED_FOR'])
    || !(in_array(@$_SERVER['REMOTE_ADDR'], ['127.0.0.1', 'fe80::1', '::1']) || php_sapi_name() === 'cli-server')
) {
    header('HTTP/1.0 403 Forbidden');
    exit('You are not allowed to access this file. Check '.basename(__FILE__).' for more information.');
}
```

tmatsuo:
Thanks, will use php_sapi_name

tmatsuo:
Ok, I just changed it to php_sapi_name(), PTAL

bshaffer:
we should also add `in_array(@$_SERVER['REMOTE_ADDR'], ['127.0.0.1', 'fe80::1', '::1'])`. There are many people who set up apache servers (or other web servers) to run locally.

tmatsuo:
Ok, added that check, PTAL

bshaffer:
sorry to be picky here. We can save ourself a line by doing this:

```php
// Turn on debug locally
if (in_array($_SERVER['REMOTE_ADDR'], ['127.0.0.1', 'fe80::1', '::1'])
    || php_sapi_name() === 'cli-server')
) {
    $app['debug'] = true;
} else {
    $app['debug'] = filter_var(getenv('BOOKSHELF_DEBUG', FILTER_VALIDATE_BOOLEAN));
}
```

tmatsuo:
Thanks PTAL

bshaffer:
LGTM!

PR #124, opened 2016-05-04
Change runtime to php

bshaffer:
@tmatsuo I am not sure why the differences between `BookshelfTest` and `E2eTest` are showing up here

### DIFF
--- a/1-hello-world/app.yaml
+++ b/1-hello-world/app.yaml
@@ -1,5 +1,5 @@
 runtime: php
 vm: true
 
-env_variables:
-  DOCUMENT_ROOT: /app/web
+runtime_config:
+  document_root: web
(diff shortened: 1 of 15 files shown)


1-hello-world/app.yaml
```diff
@@ -1,2 +1,5 @@
-runtime: custom
+runtime: php
 vm: true
+
+env_variables:
```
tmatsuo: You should use `runtime_config` instead of `env_variables`.

```yaml
runtime:php
vm:true

runtime_config:
  document_root: web
```

tmatsuo:
LGTM

tmatsuo:
Note: we should republish our docs

PR #129, opened 2016-06-14
adds support for cloud SQL second gen

bshaffer:


2-structured-data/src/DataModel/CloudSql.php
```diff
@@ -53,7 +53,18 @@ public function __construct($dsn, $user, $password)
             return explode(' ', $columnDefinition)[0];
         }, $columns);
         $columnText = implode(', ', $columns);
-        $pdo = $this->newConnection();
+
+        try {
+            $pdo = $this->newConnection();
+        } catch (\PDOException $e) {
+            if (false === strpos($e->getMessage(), 'Unknown database')) {
```
tmatsuo: I wish there was a better way to do this, but presumably it is the best way?

bshaffer: No, if we tell the users how to create the DB in the tutorial using the cloudSQL proxy, this can be removed.

bshaffer:
@tmatsuo made some pretty large changes - removed `google/apiclient` in favor of `DatastoreClient`. PTAL!

tmatsuo:
The test says "Class 'Google_Client' not found in /home/travis/build/GoogleCloudPlatform/getting-started-php/4-auth/src/app.php on line 72"

2-structured-data/src/DataModel/Datastore.php
```diff
@@ -36,110 +39,60 @@ class Datastore implements DataModelInterface
...
-        $key = $response->getMutationResults()[0]->getKey();
+        $key = $this->datastore->key('Book');
+        $entity = $this->datastore->entity($key, $book);
 
+        $this->datastore->insert($entity);
 
-        // return the ID of the created datastore item
-        return $key->getPath()[0]->getId();
+        // return the ID of the created datastore entity
+        return $entity->key()->pathEnd()['id'];
```
tmatsuo: Now the `Key` has `pathEndIdentifier` method, maybe you can use it instead

2-structured-data/src/DataModel/Datastore.php
```diff
@@ -151,106 +104,37 @@ public function update($book)
...
-                    'update' => $entity
-                ]
-            ]
-        ]);
-
-        $response = $this->datastore->projects->commit($this->datasetId, $request);
+        $transaction = $this->datastore->transaction();
+        $key = $this->datastore->key('Book', $book['id']);
+        $task = $transaction->lookup($key);
+        unset($book['id']);
```
tmatsuo: Why do you `unset` the `id`?

2-structured-data/src/DataModel/Datastore.php
```diff
@@ -151,106 +104,37 @@ public function update($book)
...
-
-        return $key;
+        $key = $this->datastore->key('Book', $id);
+        return $this->datastore->delete($key);
     }
 
     private function verifyBook($book)
     {
+        if ($book instanceof Entity) {
+            $book = $book->get();
```
tmatsuo: Why do you need to call `get`? How about to add a comment?

2-structured-data/src/DataModel/Datastore.php
bshaffer: The "data model" interface expects an array with the "id" property set. But if you save the model with that property, the Datastore API throws an error.

2-structured-data/src/DataModel/Datastore.php
bshaffer: I'll look into cleaning this up

bshaffer: removed

tmatsuo: I still see `$book->get();`

2-structured-data/src/DataModel/Datastore.php
bshaffer: will do!

2-structured-data/src/DataModel/CloudSql.php
bshaffer: We now cover this step in the tutorial, so we can remove this

2-structured-data/src/DataModel/Datastore.php
```diff
@@ -36,110 +39,60 @@ class Datastore implements DataModelInterface
...
+        $results = $this->datastore->runQuery($query);
 
         $books = [];
-        foreach ($batch->getEntityResults() as $entityResult) {
-            $entity = $entityResult->getEntity();
-            $book = $this->propertiesToBook($entity->getProperties());
-            $book['id'] = $entity->getKey()->getPath()[0]->getId();
+        $nextPageCursor = null;
+        foreach ($results as $entity) {
+            $book = $entity->get();
```
tmatsuo: Do you need to call `->get()` here as well?

bshaffer: yes because the interface expects an array to be returned

PR #131, opened 2016-06-16
adds pubsub

bshaffer:


6-pubsub/composer.json
```diff
@@ -7,10 +7,12 @@
   "require": {
     "php": "~5.5.0|~5.6.0|~7.0.0",
     "google/apiclient": "~2.0@dev",
```
tmatsuo: Shouldn't it be "google/apiclient": "~2.0"?

bshaffer: yes, I can make this change in a separate PR

bshaffer: So, `google/cloud` only supports `0.7` of `google/auth`, and `google/apiclient` supports `0.8`. However, the master branches of *both* support `0.9`, so I think it's best to wait until the next versions are tagged to change this.

tmatsuo: That's fine

2-structured-data/src/app.php
```diff
@@ -40,6 +43,35 @@
 
 $app['config'] = Yaml::parse(file_get_contents($config));
 
+// determine the datamodel backend using the app configuration
+$app['bookshelf.model'] = function ($app) {
+    /** @var array $config */
+    $config = $app['config'];
+
+    // Data Model
+    switch ($config['bookshelf_backend']) {
```
tmatsuo: Nit: not a big deal, but it will throw Undefined index.

6-pubsub/src/PubSub/Worker.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+namespace Google\Cloud\Samples\Bookshelf\PubSub;
+
+use Google\Cloud\PubSub\Subscription;
+
+class Worker
+{
+    const TOPIC_NAME = 'fill-book-details';
+    const SUB_NAME = 'book-worker-sub';
```
tmatsuo: Nit: What about moving them to yaml files or something? Not a big deal, but it will give us flexibility for tests.

bshaffer: I like this idea!

6-pubsub/src/PubSub/entrypoint.php
```diff
@@ -0,0 +1,67 @@
...
+$subscription = $topic->subscription(Worker::SUB_NAME);
+if (!$subscription->exists()) {
+    $subscription->create();
+}
+
+// add our job to the event loop
+$job = new LookupBookDetailsJob($app['bookshelf.model'], $app['google_client']);
+$worker = new Worker($subscription, $job);
+$server->loop->addPeriodicTimer(0, $worker);
+$server->run();
```
tmatsuo: Showing how to use Ratchet is nice, but I also feel like it's too complicated for this use case.

General questions:
- Is this already working on prod (Our docker image not yet has php in its path)?
- Is it stable enough?
- Ratchet has a dependency for an old guzzle. Isn't it a big issue?
- How scalable it is?

Have you considered other approaches, like using pthreads, and custom supervisord.conf, etc?

bshaffer: Good questions.  Here are some answers

- yes, it is working in prod. I just had to add a custom `Dockerfile` to test.
- Ratchet is very stable
- Fortunately, the older version of guzzle does not conflict, as it's a different namespace.
- I have not tested this, although it's modeled off the other language worker patterns, so I believe it would scale very well. Multiple instances of the "worker" service can be spun up to handle larger loads if necessary.

tmatsuo: Alright, LGTM once the test passes

bshaffer: @tmatsuo `pthreads` looks very promising, although it's only for php7. I am OK with forcing PHP7 in the cloud, but it may be a little early to force that locally.

6-pubsub/worker.yaml
```diff
@@ -0,0 +1,8 @@
+#runtime: php
+vm: true
+service: worker
+entrypoint: php src/PubSub/entrypoint.php
```
tmatsuo: Nit: This means no cron daemon will run. The log file may become big. or is it now handled by the host machine now?

bshaffer: I believe this is handled by the host machine, but we should confirm this.

2-structured-data/test/e2e/CloudSqlTest.php
```diff
@@ -22,9 +22,9 @@
  */
 class CloudSqlTest extends E2eTest
 {
-    protected static function setBackEnd()
+    protected static function getCustomConfig()
     {
-        putenv('BOOKSHELF_BACKEND=cloudsql');
+        return ['bookshelf_backend' => 'cloudsql'];
```
bshaffer: @tmatsuo please note the change in how e2e tests work - the setting of the environment variables was having an adverse effect on the tests.

tmatsuo:
I still prefer using supervisord, but it is just a matter of preference. LGTM

Please file an issue for investigating on the log rotation issue.

SurferJeffAtGoogle:
LGTM.

PR #132, opened 2016-06-16
move bookshelf.model service into app.php

bshaffer:


bshaffer:
Breaking the pubsub PR into smaller parts...

2-structured-data/src/app.php
```diff
@@ -40,6 +43,35 @@
 
 $app['config'] = Yaml::parse(file_get_contents($config));
 
+// determine the datamodel backend using the app configuration
+$app['bookshelf.model'] = function ($app) {
+    /** @var array $config */
+    $config = $app['config'];
+
+    // Data Model
+    switch ($config['bookshelf_backend']) {
```
tmatsuo: Nit: it will throw Undefined Index when 'bookshelf_backend' is missing. It might not be a big deal since we have config template, but it is still possible you accidentally change that key (I have done such things many times).

bshaffer:
@tmatsuo added tests for coverals - PTAL!

tmatsuo:
Overall looks good with one nit

2-structured-data/src/app.php
bshaffer: TBH I really don't want to surround this (already verbose) datamodel switcher with an `isset`. If we have it set in the configs, and someone makes a typo and removes it, the warning (an exception) they get will be enough to help them debug.

bshaffer: I could add this block, if that would be a good compromise:

```php
if (empty($config['bookshelf_backend'])) {
    throw new \DomainException('"bookshelf_backend" must be set in config/settings.yml');
}
```

tmatsuo: Yeah, I prefer adding something like that!

bshaffer: done.

tmatsuo:
LGTM

PR #137, opened 2016-07-06
updates google apiclient to stable version

bshaffer:
upgrades `google/cloud` to `v0.4` and `google/client` to `^2.0`


tmatsuo:
LGTM

tmatsuo:
Looks like we need this for the test builds, merging first.

PR #140, opened 2016-07-08
Added envvars for e2e tests.

tmatsuo:
Fixing pubsub e2e test.

### DIFF
--- a/_tuts/6-pubsub.diff
+++ b/_tuts/6-pubsub.diff
@@ -979,6 +979,19 @@ index c2e1dbb..104bb5d 100644
          return $app->redirect("/books/$id");
      }
  
+diff --git a/test/app-e2e.yaml b/test/app-e2e.yaml
+index ae1a580..82b35ee 100644
+--- a/test/app-e2e.yaml
++++ b/test/app-e2e.yaml
+@@ -3,6 +3,8 @@ vm: true
+ 
+ env_variables:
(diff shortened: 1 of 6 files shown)


tmatsuo:
Also running the full tests on our jenkins

tmatsuo:
Seems like it needs to be `runtime: custom` for container engine e2e test.
Why it's deploying to App Engine flexible BTW?

tmatsuo:
Now the e2e tests are passing on our jenkins :)

_tuts/6-pubsub.diff
```diff
@@ -979,6 +979,89 @@ index c2e1dbb..104bb5d 100644
...
+   BOOKSHELF_CONFIG: /app/config/settings-e2e.yml
++  NGINX_CONF_INCLUDE: config/nginx-app.conf
++  PHP_INI_OVERRIDE: config/php-app.ini
+ 
+ runtime_config:
+   document_root: web
+diff --git a/test/e2e/BookshelfTest.php b/test/e2e/BookshelfTest.php
+deleted file mode 100644
+index 769e07d..0000000
+--- a/test/e2e/BookshelfTest.php
```
bshaffer: Something is wrong with this - this file should not be affected by this change

Can you do a `git add -i` and not add this hunk? We don't want this to keep showing up in PRs.

bshaffer:
LGTM

PR #145, opened 2016-08-23
fix cs for pubsub worker test

bshaffer:


bshaffer:
closed in favor of #144

PR #155, opened 2017-03-14
Update gcloud commands in GCE deployment

jonparrott:
Context https://github.com/GoogleCloudPlatform/golang-samples/pull/205


bshaffer:
Thank you for this! Unfortunately our tutorial-building tool isn't sophisticated enough to take pull requests (it's a feature we're working on) so I had to create a new PR. See #156

PR #158, opened 2017-05-15
Fix Worker to submit valid messages for acknowledgeBatch

beastafk:
The example was broken because the 
```
$subscription->acknowledgeBatch()
```
method expects to receive an array of `Google\Cloud\PubSub\Message` objects, not just the Ids.


bshaffer:
fixed in #163 
Thanks @beastafk

PR #159, opened 2017-06-26
First draft of Bookshelf Postgres Updates

ryanmats:


2-structured-data/config/settings.yml.dist
```diff
@@ -2,15 +2,22 @@
...
 
-# options are "cloudsql", "mongodb", or "datastore"
-bookshelf_backend: cloudsql
+# options are "cloudsql-mysql", "cloudsql-postgres", "mongodb", or "datastore"
+bookshelf_backend: cloudsql-mysql
 
-# configure CloudSQL backend
-mysql_dsn: "mysql:host=IP_ADDRESS;port=3306;dbname=YOUR_DB_NAME"
+# configure CloudSQL for MySQL backend
+mysql_dsn_local: "mysql:host=127.0.0.1;port=3306;dbname=YOUR_DB_NAME"
```
bshaffer: This all looks great! One thing that is a little confusing is that `cloudsql-mysql` with `mysql_dsn_local` actually has nothing to do with CloudSQL.

One way we could fix this would be to rename a few things
 - `Datamodel\CloudSql => Datamodel\Sql` 
 - `cloudsql-mysql => mysql` 
 - `cloudsql-postgres => postgres` 
 - `mysql_dsn_deployed => mysql_dsn_cloudsql`

2-structured-data/config/settings.yml.dist
```diff
@@ -2,15 +2,22 @@
...
+mysql_dsn_deployed: "mysql:unix_socket=/mysql/INSTANCE_CONNECTION_NAME;dbname=YOUR_DB_NAME"
 mysql_user: YOUR_MYSQL_USER
 mysql_password: YOUR_MYSQL_PASSWORD
-cloudsql_connection_name: "INSTANCE_CONNECTION_NAME"
+
+# configure CloudSQL for Postgres backend
+postgres_dsn_local: "pgsql:host=127.0.0.1;port=5432;dbname=YOUR_DB_NAME"
+postgres_dsn_deployed: "pgsql:host=/cloudsql/INSTANCE_CONNECTION_NAME;dbname=YOUR_DB_NAME"
```
bshaffer: To simplify things further, instead of having `*_dsn_local` and `*_dsn_deployed`, just make a `sql_dbname` and automatically create the DSNs from `cloudsql_connection_name`, `bookshelf_backend`, and whether or not we're on app engine.

Thoughts?

ryanmats: this sounds good - made these changes. Now there are fewer variables to fill out in the config file.

bshaffer:
The reason the tests are failing is that [`GetConfigTrait`](https://github.com/GoogleCloudPlatform/getting-started-php/blob/master/2-structured-data/test/lib/GetConfigTrait.php#L38) is still returning `mysql_*` config values, which are being set in [`AppTest.php`](https://github.com/GoogleCloudPlatform/getting-started-php/pull/159/files#diff-08b0fd92382c980e80a5091ab02c8456R49) as those same values, so `cloudsql_*` is not being set, and the default values (e.g. `YOUR_CLOUDSQL_USERNAME`) are beign used instead.

2-structured-data/src/DataModel/Sql.php
```diff
@@ -42,7 +42,7 @@ public function __construct($dsn, $user, $password)
             'id serial PRIMARY KEY ',
             'title VARCHAR(255)',
             'author VARCHAR(255)',
-            'publishedDate VARCHAR(255)',
+            'publisheddate VARCHAR(255)',
```
bshaffer: Why are we lowercasing this? `imageUrl` below has the same casing, as does `createdBy`. It'd be best to be consistent unless we have a strong reason to change it. If we need to change it, we should change all of them to snake case (i.e. `published_date`

testing/run_test_suite.sh
```diff
@@ -20,7 +20,7 @@ MYDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 source ${MYDIR}/variables.sh
 
 # Coding style check.
-php-cs-fixer fix --diff
+#php-cs-fixer fix --diff
```
bshaffer: add this flag to `php_cs.dist`: https://github.com/GoogleCloudPlatform/php-docs-samples/pull/406/files

2-structured-data/composer.json
```diff
@@ -9,7 +9,8 @@
     "google/cloud": "^0.21",
     "silex/silex": "^1.3",
     "symfony/yaml": "^3.0",
-    "twig/twig": "^1.22"
+    "twig/twig": "^1.22",
+    "mongodb/mongodb": "^1.0.0"
```
bshaffer: I believe this addition was in error. `mongodb` is an optional dependency, and so shouldn't be included in `composer.json`.

bshaffer:
Closing, as this was merged in #162

PR #160, opened 2017-07-22
Updates Dependencies

jgacuca567:


bshaffer:
What is the justification for this change?

bshaffer:
We want to stay with PHPUnit 4 at this time, as we're still testing PHP versions 5.6, for which PHPUnit 6 is unavailable

PR #164, opened 2017-08-31
Should MySQL be included in the options?

jgacuca567:
If so, then it should be updated on the following page.  https://cloud.google.com/php/getting-started/using-cloud-sql-with-mysql

### DIFF
--- a/2-structured-data/config/settings.yml.dist
+++ b/2-structured-data/config/settings.yml.dist
@@ -3,7 +3,7 @@ google_client_id:      YOUR_CLIENT_ID
 google_client_secret:  YOUR_CLIENT_SECRET
 google_project_id:     YOUR_PROJECT_ID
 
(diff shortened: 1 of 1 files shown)


bshaffer:
I am unsure what you're referring to. "mysql" IS included in the options, and also the documentation in the page you linked to states "Set the value of bookshelf_backend to mysql."

What is the problem exactly?

jgacuca567:
@bshaffer The only options in the page are `cloudsql`, `mongodb` or `datastore`. Not **mysql** in the documentation.

bshaffer:
Ahh yes, thank you for bringing this up. It used to be "mysql" but was changed to "cloudsql" when postgres support was added to cloudsql.

bshaffer:
@jgacuca567 The option *used* to be `cloudsql`, but once postgres was added, we split this apart to be `postgres` and `mysql`. I believe the documentation and code are consistent with this.

PR #165, opened 2017-09-16
Updates Bootstrap 3.3.2 to 3.3.7

jgacuca567:


bshaffer:
Thank you for your contribution! Is there a significant fix in this change which is applicable to the bookshelf app?

jgacuca567:
@bshaffer, I don't think so, it just updating to the most current version.

PR #168, opened 2017-11-17
Keeping costs low

devlance:
We've gotten reports of unexpected costs from users that left the hello world running and had two standard instances running for a month. This keeps the costs to a minimum to kick the tires.


tmatsuo:
I believe @JustinBeckwith has an opinion on this

bshaffer:
This is great! I've updated it in favor of #171

PR #170, opened 2017-11-29
gke rename

bshaffer:
cc @ahmetb


ahmetb:
Please update `README.md` as well.

ahmetb:
You can update links to this tutorial on GKE to read like `bookshelf-on-kubernetes-engine`. They'll be broken for one or two days (here, in this repo) but we won't need to fix them again.

bshaffer:
@ahmetb done

bshaffer:
Note the tutorial URLs on cloud.google.com still have "container" in them. This is because they haven't been updated yet, and so it's best to leave them until they're fixed. And since we'll be redirecting the old URLs they won't cause any harm as-is.

ahmetb:
LGTM

PR #175, opened 2018-01-04
Update worker.yaml

jgacuca567:
According to `gcloud` Deploymnets using `vm`have been deprecated. Have replaced with `env:flex`


tmatsuo:
@jgacuca567 Thanks! Since there are other issues in `worker.yaml`, we'll fix it in another PR.

PR #178, opened 2018-01-10
changes vm:true to env:flex

bshaffer:


tmatsuo:
Actually there are other issues with this step, so the pubsub worker won't work even with this change, but LGTM

tmatsuo:
Closing, covered in #179

PR #179, opened 2018-01-10
Moved to supervisord config from the entrypoint

tmatsuo:


optional-kubernetes-engine/config/php-pubsub.ini
```diff
@@ -1,3 +0,0 @@
-extension=mongo.so
```
bshaffer: Why are removing this file? I believe we still want these overrides.

6-pubsub/worker.yaml
```diff
@@ -1,11 +1,24 @@
...
-env_variables:
-  PHP_INI_OVERRIDE: config/php-pubsub.ini
-# [END overrides]
+# [START runtime_config]
+runtime_config:
+  document_root: .
+  supervisord_conf: config/worker-supervisord.conf
+# [END runtime_config]
+
+## Uncomment for Cloud SQL
```
bshaffer: I do not believe Cloud SQL is relevant for the workers. They do not use a database. So for this reason we should get rid of this line.

optional-kubernetes-engine/config/php-pubsub.ini
tmatsuo: `mongo` is not available (let's use `mongodb`), `curl` is built-in, so not needed.

`suhosin`, yeah maybe needed for 5.6, but we pick up 7.1 runtime with the current configuration anyways, so it's not needed unless you modify the composer.json and explicitly allow only php 5.6.

6-pubsub/worker.yaml
tmatsuo: The worker is updating the thumbnail (`image_url`) field in the database. I believe it's needed.

tmatsuo:
@bshaffer 

Ok, this version is working for me locally. Also I removed the unnecessary config files from the step after pubsub. PTAL

optional-kubernetes-engine/config/worker-supervisord.conf
```diff
@@ -1,11 +0,0 @@
-[program: pubsub-worker]
```
bshaffer: Don't we want this file still for K8s Engine? It still uses the same docker image / runtime, and so the worker won't run without it.

tmatsuo: The GKE version is using the custom Dockerfile and it has `CMD` for running the worker.
Potentially we can move it to supervisord, but now I want to fix the step 6 first.

tmatsuo:
@bshaffer anything left?

PR #184, opened 2018-01-19
If we change the published_date

jgacuca567:
@tmatsuo 
Where I can place the NOW() Function?


tmatsuo:
@jgacuca567 

Thanks, but this seems to be a question rather than a PR. Can you file an issue instead?

JamesVanWaza:
Whats the status? Its been 4 months...... 

@bshaffer
@tmatsuo

I created the issue, already #186

Issue #180, opened 2018-01-10
Flaky tests


tmatsuo:
Tests are now very flaky

tmatsuo:
Most of the time, the mongodb driver throws a timeout. Maybe we can have longer timeout for tests with `connectTimeoutMS` option.
http://php.net/manual/en/mongodb-driver-manager.construct.php

tmatsuo:
Ok, the test says:
```
No suitable servers found: `serverselectiontimeoutms` timed out:
```
So longer `serverSelectionTimeoutMS`

tmatsuo:
Raised it to 120000

tmatsuo:
The tests are constantly failing with the wrong MONGO_URL envvar that I set in https://github.com/GoogleCloudPlatform/getting-started-php/issues/180#issuecomment-357403052

I just fixed the envvar and see how it goes.

tmatsuo:
Today I looked at the failed test and found the following log:

```
0.83s$ testing/test_cleanup.sh
ERROR: (gcloud.compute.firewall-rules.delete) Could not fetch resource:
 - The resource 'projects/getting-started-php/global/firewalls/travis-job-333550386' was not found
```

The real cause might be the firewall has not created properly.

Currently the test script is doing like

1) Delete the firewall || /bin/true
2) Create the firewall
3) Delete the firewall in the cleanup script

My hypothesis is that, somehow the operation for the above 1) is errornously executed after 2).

I'll experiment tomorrow.

tmatsuo:
Unfortunately, this is because our security system routinely remove the temporary firewall rule we created before the test. The internal bug: b/72542269

tmatsuo:
We switched to use the local mongodb on the travis test runner and it mitigated the flakiness

PR #185, opened 2018-01-26
Stop deleting the firewall.

tmatsuo:
See
https://github.com/GoogleCloudPlatform/getting-started-php/issues/180#issuecomment-360698613

### DIFF
--- a/testing/install_test_dependencies.sh
+++ b/testing/install_test_dependencies.sh
@@ -96,7 +96,6 @@ else
         "${GOOGLE_APPLICATION_CREDENTIALS}"
 fi
 
-gcloud compute firewall-rules delete ${FIREWALL_NAME} || /bin/true
 # Create a firewall rule for mongodb
 IP=`curl https://ip-dot-cloud-dpes.appspot.com/`
 gcloud compute firewall-rules create ${FIREWALL_NAME} \


tmatsuo:
The test with 7.1 failed. The log said:

```
gcloud compute firewall-rules create ${FIREWALL_NAME} \
    --allow tcp:27017 \
    --source-ranges ${IP}/32 \
    --target-tags mongodb \
    --description "Allow mongodb access for ${FIREWALL_NAME}"
Creating firewall...
................Created [https://www.googleapis.com/compute/v1/projects/getting-started-php/global/firewalls/travis-job-333603615].
done.
NAME                  NETWORK  DIRECTION  PRIORITY  ALLOW      DENY
travis-job-333603615  default  INGRESS    1000      tcp:27017
```
and

```
$ testing/test_cleanup.sh
ERROR: (gcloud.compute.firewall-rules.delete) Could not fetch resource:
 - The resource 'projects/getting-started-php/global/firewalls/travis-job-333603615' was not found
```
So gcloud reported that it successfully created the firewall, and then later the same firewall was not found. The name of the firewall is unique to the travis job, so I don't think it's naming conflict.

I think I may understand what's happening...

PR #191, opened 2018-05-15
Return string oid upon creating a book with the MongoDB backend

### LINKED ISSUE #190: [pubsub] Fails to create a book with the mongodb backend
The error:

Google\Cloud\Exception\BadRequestException: {   "error": {     "code": 400,     "message": "Invalid JSON payload received. Unknown name \"$oid\" at 'messages[0].attributes[0].value': Cannot find field. .....

tmatsuo:
fixes #190


tmatsuo:
Hmm, still seeing some flakes in the tests

PR #199, opened 2019-05-15
[DO NOT MERGE] Lumen WIP

bshaffer:


bookshelf/resources/views/list.php
```diff
@@ -12,48 +13,42 @@
...
 <h3>Books</h3>
 <a href="/books/add" class="btn btn-success btn-sm">
   <i class="glyphicon glyphicon-plus"></i>
   Add book
 </a>
 
-{# [START book_list] #}
-{% for book in books %}
+<?php // [START book_list] ?>
+<?php foreach ($books as $i => $book): ?>
```
grayside: It would be helpful to have a comment block at the top of each template with the expected variables described.

E.g., a list including book, it's type, short description. Like a method param doc.

bookshelf/routes/web.php
```diff
@@ -0,0 +1,199 @@
...
+$storage = new StorageClient([
+    'projectId' => $projectId,
+]);
+$bucketId = $projectId . '.appspot.com';
+$gcsBucket = $storage->bucket($bucketId);
+# [END cloud_storage_client]
+
+/*
+|--------------------------------------------------------------------------
+| Application Routes
```
grayside: A number of the application route snippets do not have much context on the variables they use. Also, more region tag namespacing.

bookshelf/test/ControllersTest.php
```diff
@@ -0,0 +1,174 @@
...
+        return require __DIR__.'/../index.php';
+    }
+
+    public static function setUpBeforeClass()
+    {
+        self::checkProjectEnvVars();
+
+        // Create a new collection for this tests so we can remove it after
+        $collectionName = 'books-' . time() . rand();
+        putenv('FIRESTORE_COLLECTION=' . $collectionName);
```
grayside: Nice trickery with putenv!

bookshelf/test/ControllersTest.php
```diff
@@ -0,0 +1,174 @@
...
+        $this->assertContains(
+            'moby-dick.png',
+            $response->getContent()
+        );
+    }
+
+    /**
+     * @depends testCreateBook
+     */
+    public function testPagination()
```
grayside: Nice, so rare to see pagination tests.

.gitignore
```diff
@@ -1,4 +1 @@
-**/vendor
```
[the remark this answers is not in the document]
bshaffer: Yes! Why wouldn't it?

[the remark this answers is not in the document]
bshaffer: This was moved into `bookshelf/.gitignore`, which I think is more appropriate

.gitignore
```diff
@@ -1,4 +1 @@
...
-composer.lock
-.gcloudignore
 .php_cs.cache
```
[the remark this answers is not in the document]
bshaffer: The `.php_cs.cache` file is  ignored though... so it should never show up in a PR

bookshelf/app/Exceptions/Handler.php
```diff
@@ -0,0 +1,53 @@
...
+     * A list of the exception types that should not be reported.
+     *
+     * @var array
+     */
+    protected $dontReport = [
+        AuthorizationException::class,
+        HttpException::class,
+        ModelNotFoundException::class,
+        ValidationException::class,
+    ];
```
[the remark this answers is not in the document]
bshaffer: These were the default values. I agree I can't think of a good reason.

bshaffer:
This PR is causing new pushes to not run in travis presumably because of a travis bug. Opening #205 instead

bookshelf/routes/web.php
```diff
@@ -0,0 +1,199 @@
...
+    if (!$snapshot->exists()) {
+        return new Response('', Response::HTTP_NOT_FOUND);
+    }
+
+    // Get book data from the request object
+    $bookData = $request->request->all();
+    $bookData['id'] = $bookId;
+
+    // [START add_image]
+    $files = $request->files;
```
dwsupplee: What do you think about turning this into a helper since the same logic lives in the `/books/add` action?

bookshelf/routes/web.php
```diff
@@ -0,0 +1,199 @@
...
+    $updateData = [];
+    foreach ($bookData as $key => $value) {
+        $updateData[] = ['path' => $key, 'value' => $value];
+    }
+
+    if ($bookRef->update($updateData)) {
+        return redirect("/books/$bookId");
+    }
+
+    return new Response('Could not update book');
```
dwsupplee: - Will this will ever be executed? It looks like either a result will be returned from `update` or an exception would be bubbled up from the transport. Perhaps we should instead be catching exceptions from `update`/ `upload` and then return this response?
- What is the flow like on the user's end if this response is returned?
- Should it return an error code?
- Why don't we return responses like this for the other actions if something fails?

bookshelf/routes/web.php
```diff
@@ -0,0 +1,199 @@
...
+        $object = $gcsBucket->upload($f, [
+            'metadata' => ['contentType' => $image->getMimeType()],
+            'predefinedAcl' => 'publicRead',
+        ]);
+        $bookData['image_url'] = $object->info()['mediaLink'];
+    }
+    // [END add_image]
+
+    // Create array to update the Firestore document
+    $updateData = [];
```
dwsupplee: You should be able to use `DocumentReference::set` here, with the `merge` option set to `true`. Then you won't need to build out this `updateData` array.

Also, I see we had validation in the old code - was there a need to remove it? Currently, it would be very easy to inject bad data into the collection.

bookshelf/test/ControllersTest.php
```diff
@@ -0,0 +1,174 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+use Google\Cloud\TestUtils\TestTrait;
+use Google\Cloud\Firestore\FirestoreClient;
```
dwsupplee: Small nit, these imports aren't ordered lexicographically.

bookshelf/test/ControllersTest.php
```diff
@@ -0,0 +1,174 @@
...
+        $this->assertContains(
+            'The Cat in the Hat (edited)',
+            $response->getContent()
+        );
+        $this->assertContains('**New Description**', $response->getContent());
+    }
+
+    public function testLogging()
+    {
+        // TODO
```
dwsupplee: Friendly reminder on this guy :)

bookshelf/app/Exceptions/Handler.php
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * Render an exception into an HTTP response.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Exception  $exception
+     * @return \Illuminate\Http\Response|\Illuminate\Http\JsonResponse
+     */
+    public function render($request, Exception $exception)
+    {
+        return parent::render($request, $exception);
```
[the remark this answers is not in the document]
bshaffer: The reasoning here is this file is generated by Lumen as an example so I left it in. I agree it's more confusing than helpful to leave it in, however.

bookshelf/composer.json
```diff
@@ -0,0 +1,19 @@
+{
+  "require": {
+    "laravel/lumen-framework": "5.8.*",
+    "monolog/monolog": "^1.24",
+    "google/cloud-storage": "^1.11",
+    "google/cloud-firestore": "^1.2",
+    "google/cloud-logging": "^1.14",
+    "google/cloud-error-reporting": "^0.12.5"
```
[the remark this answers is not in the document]
bshaffer: That would be great! It also serves as a reminder for me to add the auto-update dependency job to this repo!

bookshelf/resources/views/form.php
```diff
@@ -12,50 +13,47 @@
...
 # limitations under the License.
-#}
-
-{# [START book_form] #}
-{% extends "base.html.twig" %}
+ob_start() ?>
 
-{% block content %}
-<h3>{{action}} book</h3>
+<?php // [START book_form] ?>
```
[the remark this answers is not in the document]
bshaffer: removed all templating region tags - we don't display these in the docs any more

bookshelf/resources/views/list.php
bshaffer: Great idea! Added.

bookshelf/resources/views/list.php
```diff
@@ -12,48 +13,42 @@
...
-      {% if book.image_url %}
-        <img src="{{book.image_url}}">
-      {% else %}
-        <img src="http://placekitten.com/g/128/192">
-      {% endif %}
-    </div>
-    {# [END book_image] #}
+  <a href="/books/<?= $book->id() ?>">
+    <?php // [START book_image] ?>
+    <?php if (!empty($book['image_url'])): ?>
```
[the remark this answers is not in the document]
bshaffer: Yes. I did it because when creating a new book, there's no way to do an empty snapshot. I agree it was a bit confusing, so now, everything uses `->get('field')` for custom fields, `->id()` for ID (because that's a special field), and `$book ? $book->get('field') : ''` for the form values.

bookshelf/resources/views/show.php
```diff
@@ -32,23 +28,24 @@
...
-{# [START book_details] #}
+<?php // [START book_details] ?>
 <div class="media">
-  {# [START book_image] #}
+  <?php // [START book_image] ?>
+  <?php if ($imgUrl = $book->get('image_url')): ?>
   <div class="media-left">
-    <img class="book-image"
-         src="{{ book.image_url ?: 'http://placekitten.com/g/128/192' }}">
+    <img class="book-image" src="<?= $imgUrl ?>" />
```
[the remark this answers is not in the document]
bshaffer: `<?=` is for output, so unless there's something being printed I use `<?php`. The comments could go either way but I think it makes more sense to keep them as `<?php` since nothing is being output there.

bookshelf/routes/web.php
bshaffer: I am not opposed to adding a helper function but I thought in this case, not being DRY would make the code easier to follow.

bookshelf/routes/web.php
bshaffer: Great suggestion re: merge=true. Added.

Great suggestion re: valudation. Added back.

bookshelf/routes/web.php
bshaffer: I wasn't sure if there was a case where `update` would just return `false` with no Exception. For this app, it's sufficient for the exception to be thrown. Other points are solid. Removed.

bookshelf/test/ControllersTest.php
bshaffer: No, they're in christmas tree order!

fixed 😄

bookshelf/test/ControllersTest.php
bshaffer: Is this sarcasm??

bookshelf/test/ControllersTest.php
bshaffer: Done! Added both functional and E2E tests!

bookshelf/resources/views/list.php
```diff
@@ -12,48 +13,42 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-#}
-
-{% extends "base.html.twig" %}
-
-{% block content %}
+ob_start()
```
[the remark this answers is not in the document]
bshaffer: Was hoping the templates would widely be ignored since that's not where any real interesting logic is, but a comment will do nicely!

bookshelf/app/Exceptions/Handler.php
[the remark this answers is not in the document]
bshaffer: Removed

.travis.yml
```diff
@@ -30,24 +31,27 @@ branches:
...
+- $COMPOSER_BIN_DIR/dump_credentials.php
+- $COMPOSER_BIN_DIR/install_test_deps.sh --gcloud --cs-fixer
 
 script:
+# Run linter
 - php-cs-fixer fix --diff
-- php vendor/bin/phpunit
+# Run all tests in all sample directories
+- |
+  set -e
```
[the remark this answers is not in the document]
bshaffer: removed this in favor of just explicitly calling bookshelf tests

PR #208, opened 2019-08-19
Getting Started: GCE

bshaffer:


gce/scripts/startup-script.sh
```diff
@@ -0,0 +1,58 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# [START getting_started_gce_startup_script]
+set -e
+export HOME=/root
+
+# Install PHP and dependencies from apt
+apt-get update
+apt-get install -y git nginx php7.2 php7.2-fpm php7.2-mysql php7.2-dev \
```
grayside: Should this be on PHP 7.3?

gce/test/DeployTest.php
```diff
@@ -0,0 +1,105 @@
...
+ */
+
+namespace Google\Cloud\GettingStarted;
+
+use Google\Cloud\TestUtils\FileUtil;
+use Google\Cloud\TestUtils\DeploymentTrait;
+use Google\Cloud\TestUtils\EventuallyConsistentTestTrait;
+use Google\Cloud\TestUtils\TestTrait;
+use Google\Cloud\Utils\ExponentialBackoff;
+use Google\Cloud\Logging\LoggingClient;
```
grayside: If the fluentd agent is present, what is the logging client needed for?

gce/test/DeployTest.php
```diff
@@ -0,0 +1,105 @@
...
+use Google\Cloud\ErrorReporting\V1beta1\ErrorStatsServiceClient;
+use Google\Cloud\ErrorReporting\V1beta1\QueryTimeRange;
+use Google\Cloud\ErrorReporting\V1beta1\QueryTimeRange\Period;
```
grayside: Unused?

gce/scripts/startup-script.sh
bshaffer: PHP 7.3 is not available as an `apt-get` package on this version of ubuntu

gce/test/DeployTest.php
bshaffer: This is unused. Removed!

authenticate-users/index.php
```diff
@@ -94,7 +94,7 @@ function validate_assertion(string $assertion, string $certs, string $audience)
             printf("<h1>Hello %s</h1>", $email);
         }
         break;
-    case ''; break; // Nothing to do, we're running our tests
+    case '': break; // Nothing to do, we're running our tests
```
grayside: non-blocking: Surprised this clears style check.

.travis.yml
```diff
@@ -46,6 +46,6 @@ install:
 
 script:
 # Run linter
-- if [ "$RUN_CS_FIXER" = "true" ]; then ./php-cs-fixer fix --diff; fi
+- if [ "$RUN_CS_FIXER" = "true" ]; then ./php-cs-fixer fix --dry-run --diff; fi
```
grayside: More concise option to consider:
```suggestion
- test "$RUN_CS_FIXER" == "true" && ./php-cs-fixer fix --dry-run --diff
```

gce/test/DeployTest.php
```diff
@@ -0,0 +1,101 @@
...
+        $backoff->execute(function () {
+            $cmd = sprintf(
+                'gcloud compute instances get-serial-port-output %s 2>&1',
+                self::$instanceName
+            );
+            exec($cmd, $output);
+            $output = implode("\n", $output);
+            if (false === strpos($output, 'Finished running startup scripts')) {
+                echo "Waiting for startup script to complete...\n";
+                throw new \Exception('Startup script has not completed');
```
grayside: non-blocking: Should this be a little more clear?
```suggestion
                throw new \Exception('Startup script has failed');
```

.travis.yml
bshaffer: I like it

gce/test/DeployTest.php
bshaffer: No, it means it's still running. I changed it to "startup script is still running" for clarity.

authenticate-users/index.php
bshaffer: It doesn't, I found a bug where the cs check wasn't failing the test when it failed (it's now fixed).

gce/index.php
```diff
@@ -0,0 +1,67 @@
...
+|--------------------------------------------------------------------------
+|
+| Next we will include the routes file so that they can all be added to
+| the application. This will provide all of the URLs the application
+| can respond to, as well as the controllers that may handle them.
+|
+*/
+
+$app->router->group([
+    'namespace' => 'App\Http\Controllers',
```
dwsupplee: Since the sample doesn't make use of a controller, do we need to register this?

gce/phpunit.xml.dist
```diff
@@ -0,0 +1,26 @@
...
+  limitations under the License.
+-->
+<phpunit bootstrap="vendor/autoload.php">
+    <testsuites>
+        <testsuite name="PHP Getting Started GCE Test Suite">
+            <directory>test</directory>
+        </testsuite>
+    </testsuites>
+    <logging>
+        <log type="coverage-clover" target="build/logs/clover.xml"/>
```
dwsupplee: Where are we using the coverage log? Wondering it we need this in the process.

gce/composer.lock
```diff
@@ -0,0 +1,3096 @@
+{
```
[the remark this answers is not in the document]
bshaffer: I was wondering if you would notice this :) 

If you deploy an `f1-micro`, there isn't enough memory for composer to install without a lock file. This tutorial uses a `g1-small`, which does deploy without one, so we could potentially drop it.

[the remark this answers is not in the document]
dwsupplee: That's really interesting! I wouldn't have guessed the lockfile would affect the memory footprint. Given the context, I still lean removing it - but trust your decision whatever it may be.

gce/index.php
bshaffer: I believe it still needs to be registered in order to serve the helloworld

gce/phpunit.xml.dist
bshaffer: Good call, removed!

gce/index.php
dwsupplee: I was curious what would happen myself - removed it (replaced with an empty array) and it worked just fine.

bshaffer: Good catch! I removed it!

gce/composer.lock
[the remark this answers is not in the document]
bshaffer: OK I removed it!

PR #254, opened 2022-09-01
update php version to support Cloud Run buildpack deployment

### LINKED ISSUE #253: feature: support cloud run deployment for bookshelf demo app
To support deployment to Cloud Run, let's get a Dockerfile for the bookshelf app OR allow deploying as buildpack. 

Reference from other getting started apps (added Dockerfile):
- [Node.js](https://github.com/GoogleCloudPlatform/nodejs-getting-started/pull/551) 
- [Python](https://github.com/GoogleCloudPlatform/getting-started-python/pull/486)

rogerthatdev:
Deploying bookshelf app as buildpack to Cloud Run:
```gcloud run deploy bookshelf --region us-central1 --allow-unauthenticated --set-env-vars="GOOGLE_CLOUD_PROJECT=${PROJECT}" --source .```

the following error occurs in the build log: 
```laravel/lumen-framework[v5.8.0, ..., v5.8.13] require php ^7.1.3 ```

This PR changes the php version in order to circumvent.  Fixes #253

### DIFF
--- a/bookshelf/composer.json
+++ b/bookshelf/composer.json
@@ -1,6 +1,6 @@
 {
   "require": {
-    "php": ">=7",
+    "php": "^7.1.3",
     "laravel/lumen-framework": "5.8.*",
     "google/cloud-storage": "^1.11",
     "google/cloud-firestore": "^1.2",

### CODE REVIEW

**Review by iennae — APPROVED** [review]
LGTM; I questioned whether we should make php version so restrictive, but as it's only impacting the bookshelf app seems ok to me.

File: bookshelf/composer.json [hunk 1]
```diff
@@ -1,6 +1,6 @@
 {
   "require": {
-    "php": ">=7",
+    "php": "^7.1.3",
```
grayside, line 3: ^7.1.3 is the equivalent to: >= 7.1.3 < 8.0.0
That makes sense to me.

grayside: What makes less sense: why does specifying 7.1.3 in the composer.json file help? Does the buildpack deliberately setup the minimum compatible PHP version?

grayside: Async: @rogerthatdev says this is indeed end-to-end tested as working with the PHP buildpack. I'll have to learn more about how this works later, fixing the bug++

grayside (approved):

PR #256, opened 2022-11-04
Use binary-only images for installing Composer

Wirone:
See: https://blog.codito.dev/2022/11/composer-binary-only-docker-images/


Wirone:
@bshaffer I believe Kokoro CI reacts to some specific changes in the repo, so the workflow is stuck here. How can we proceed with this PR?

Wirone:
I am going to close this, because of lack of feedback from the maintainers. I don't want this PR stuck on my list, I can always reopen it it there's an interest and will to merge it.

